PR #3, opened 2015-06-23
Fixed some UI that not correct

guzzilar:
## PR Purpose
- Added link that reference to our Omise Dashboard site
- Fixed some UI that was not corrected in Dashboard page.
- Fixed about transfer box that was not shown when an account doesn't have any transfer history.
- Updated CHANGELOG.md


fred:
:+1:

PR #19, opened 2015-11-15
Updated some part of code to support php v5.2 - 5.3

guzzilar:
## PR Purpose
- Updated some part of code to support php v5.2 - 5.3.
- Add autocomplete attribute into card number and security code fields of the checkout form.


guzzilar:
:+1:

PR #22, opened 2016-01-05
Check authorized and captured status before continue the charge action

guzzilar:
## PR Purpose

- Check authorized and captured status before continue the charge action.
- Disable submit button action when user clicks submit.
- Extends `Mage_Payment_Model_Method_Abstract` class instead of `Mage_Payment_Model_Method_Cc`


guzzilar:
:+1:

Issue #26, opened 2016-09-01
Need Omise's keys dialogue in checkout


mr6502:
We are running Magento 1.9.2.4 using a custom checkout (Firecheckout module).  We've installed omise-magento 1.9.0.6 which says this issue is fixed.  As a dependency we have omise-php 2.5.0 (Side note:  The omise-magento readme doc says it requires 2.4.0 but we couldn't get it to work with 2.4.0 either).

![spectacle bh4872](https://cloud.githubusercontent.com/assets/21123937/18174528/c133eb7a-7065-11e6-974d-aad83a8c0e97.png)

Please can you suggest how to investigate or fix for this?

guzzilar:
@mr6502 Hi,

Currently, we are investigating this issue.
From what we know right now, it could be possible that our extension doesn't support with Firecheckout module.

However, we will keep inform you when we find a better solution to solve this issue.

And, thanks for report us :D
Nam

guzzilar:
Hi,

We just updated [README.md](https://github.com/omise/omise-magento/blob/master/README.md) to explain about using Omise Magento with One Page Checkout extension.

You could read at the topic "Using omise-magento with One Page Checkout (OPC) extension" from the link below,
https://github.com/omise/omise-magento/blob/master/README.md#using-omise-magento-with-one-page-checkout-opc-extension

Nam

guzzilar:
@mr6502 Hi, just wanted to give you an update, the solution has been provided at release [v1.12](https://github.com/omise/omise-magento/releases/tag/v1.12).

Thanks for reported the issue :)
Nam.

PR #27, opened 2016-09-05
Add the description for using with One Page Checkout (OPC) extension

nimid:
**1. Objective reason**

According to the problem that the payer has input the card information correctly but the system display the popup with the error message "Need Omise's keys", the cause of the problem is omise-magento has not been support or work with the extension One Page Checkout (OPC) that the merchant installed.

Most of the OPC extensions have no any procedure to trigger the omise-magento to create the Omise token normally.

Currently, to use omise-magento with OPC, the merchant need to modify their extension by themselves.

Related ticket: [Issue #26 ](https://github.com/omise/omise-magento/issues/26), T1000

**2. Description of change**

Add the description at the README.md for using omise-magento with the extension, One Page Checkout or One Step Checkout.

The source code has not been altered or modified.

**3. Users affected by the change**

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

Merchant use the normal checkout process without OPC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,102 @@ Then click the **Continue** button.
 
 6. Now head back to your admin dashboard orders section: `Sales` > `Orders`. You will see your order with `Pending` or `Processing` status (depends on your configuration).  
   ![Admin's order page](https://cdn.omise.co/assets/omise-magento/checkout-with-omise-05.png)  
+
+## Using omise-magento with One Page Checkout (OPC) extension
+By default, the checkout process of Magento has many steps and it have been displayed step by step. This default checkout process can be adjusted by the merchant.
+
+One of the adjustment is the merchant will find and install an extension to reduce the checkout steps or change the display of checkout process to be in one page.
+
+There are so many third-party extensions that delivered this feature. But most of those extensions are not compatible with omise-magento. Because omise-magento has a secure step at the client side (the JavaScript at the web browser of the payer) to create a token to be the representative of the card information of the payer. That token is used to be the representative between merchant server and Omise API for process the payment.
+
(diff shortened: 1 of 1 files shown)


guzzilar:
LGTM 👍 
btw, I think we can introduce/ have ref link to IWD One Page Checkout extension at the topic "Disclaimer about the example of third-party extension".

guzzilar:
is that https://www.iwdagency.com/extensions/one-step-page-checkout.html ?

nimid:
@guzzilar That is a good idea but I not think so.

Adding a link to the external third-party, can be increase more implication that we introduce or recommend that extension.

Although, we use that extension to be the example in this guideline but currently, we do not support any OPC. If the customer want to use the OPC, they can find and do it by themselves.

This guideline has provided the name and version of that extension. It is enough referenced information for the customer.

guzzilar:
@nimid I just don't get the downside of introduce / recommend (ok,understand that no need to say we recommend, but just introduce) that extension here, if that one can solve the issue (as a workaround method) our support team gonna introduce that extension as an alternative solution to merchants anyway. 

could you tell me more your concerned about showing the ref link in the README file?

guzzilar:
ah, my point is, I didn't mean, we gonna say "use it, it'd solve your issue"
but it's just convenience thing to help merchant to find the link that we're mentioned there.

nimid:
@guzzilar Yes, I agree with the benefit of convenient to the customer.

guzzilar:
noted, moved to #28
to merge it directly to `master` branch instead.

PR #28, opened 2016-09-13
Add the description for using with One Page Checkout (OPC) extension

guzzilar:
> noted, moved from #27

**1. Objective reason**

According to the problem that the payer has input the card information correctly but the system display the popup with the error message "Need Omise's keys", the cause of the problem is omise-magento has not been support or work with the extension One Page Checkout (OPC) that the merchant installed.

Most of the OPC extensions have no any procedure to trigger the omise-magento to create the Omise token normally.

Currently, to use omise-magento with OPC, the merchant need to modify their extension by themselves.

Related ticket: [Issue #26 ](https://github.com/omise/omise-magento/issues/26), T1000

**2. Description of change**

Add the description at the README.md for using omise-magento with the extension, One Page Checkout or One Step Checkout.

The source code has not been altered or modified.

**3. Users affected by the change**

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

Merchant use the normal checkout process without OPC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,102 @@ Then click the **Continue** button.
 
 6. Now head back to your admin dashboard orders section: `Sales` > `Orders`. You will see your order with `Pending` or `Processing` status (depends on your configuration).  
   ![Admin's order page](https://cdn.omise.co/assets/omise-magento/checkout-with-omise-05.png)  
+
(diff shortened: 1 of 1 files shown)


guzzilar:
👍

PR #38, opened 2017-01-02
Introduce 3-D Secure payment feature

guzzilar:
> This PR requires PR #39 to be merged first.

#### 1. Objective

To support 3-D Secure feature.
Merchant will be able to enable 3-D Secure config and process a 3-D Secure payment through the plugin.

> This changes support for payment "Authorize only" & "Authorize and Capture" actions.

**Related information**:
Related issue(s): 🙅
Related ticket(s): T1345

---

#### 2. Description of change

1. Implement `Gateway/Request/ThreeDSecureDataBuilder.php` to determine whether attach `process_3dsecure` and `return_uri` params to a `requestBuilder` when build a request object if 3-D Secure config is enabled.
2. Pass `return_uri` param to `Authorize` & `Capture` gateway commands if 3-D Secure config is enabled.
3. Bypass `Gateway\Validator` (for both `Authirze` and `Capture` validators) if 3-D Secure config is enabled.
4. Implement `Gateway/Response/PaymentDetailsHandler.php` to attach `OmiseCharge.id` back to a payment object after make a request to Omise API.
5. Update `di.xml` according to the changes.
6. Add new route `omise/callback/validate3dsecure` as a callback url that 3-D Secure Authorization page will redirect back after 3-D Secure Cauthen process to validate a process.
7. Implement `Controller/Callback/Validate3DSecure.php` to handle callback url (to validate 3-D Secure result).
8. Implement REST API `GET V1/order/:orderId/payment-authorize-uri` to retrieve `authorize_uri` after place an order.

Long description for number 8, I chose this way because:
- It need to use `OmiseCharge.authorize_uri` to redirect buyer to a 3-D Secure authorization page.
- But, because checkout form proceed a `placeOrder` process by calling REST API (through AJAX)
- And in the middle of that `placeOrder` API process, they call to `GatewayCommand` class.
- And then, return `orderId` back to checkout form.

```
Checkout form ============================================================= continue process ===
||                                                                   ^^ Return orderId here
|| AJAX request to REST API                                          ||
||                                                                   ||
vv                                                                   ||
•• REST API Do its functions                                         ||
    •• 1. Initiate classes or do stuff.                              ||
    •• 2. In the middle of the process, call GatewayCommand here     ||
    •• 3. GatewayCommand only "connet to 3rd-party payment gateway"  ||
    •• 4. After process payment, do "placeOrder" or something here   ||
||                                                                   ||
============================================ >>> ======================
```

- So, the payment process is at number "3".
[...]

PR #39, opened 2017-01-03
Separate OmiseConfig class out from Ui ConfigProvider class

guzzilar:
#### 1. Objective

To completely separate the responsibility of these 2 classes, and make it more sense to reuse the code.

Since `Model\Ui\OmiseConfigProvider` class only be used to provide Omise config to the checkout form (frontend) only. It should only focus to function on about `gathering data` and `return data to a form`.

And let `Model\OmiseConfig` class to communicate with Magento's database & setting to provide Omise configuration.

> Note, you can see the use case/ benefit of `Model\OmiseConfig` from PR #38 as it's being used this implementation.

**Related information**:
Related issue(s): 🙅
Related ticket(s): T2553

#### 2. Description of change

- Move functions that were implemented to retrieve Omise config out from `Model\Ui\OmiseConfigProvider` to `Model\OmiseConfig`
- `Model\Ui\OmiseConfigProvider` will only focus to function about "provide information/ configurations to a checkout form"
- Update code according to the changes (`etc/di.xml`, `Gateway/Http/Client/AbstractOmiseClient.php`)
- Add unit test.

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 2.1.2.
- **PHP version**: 7.0.14.

**✏️ Details:**

To make sure that this changes will not impact to the core functions.
Since `di.xml` & `AbstractOmiseClient.php` were updated to comply with the changes, we should consider to test 2 parts.

1. Test checkout item (either `authorise only` or `authorise & capture` are fine to test)
The step are as follows:
1.1. Add item to a cart & checkout as normal process.
1.2. Test checkout a cart with `Omise Payment` payment method, it should success.
1.3. Then, go to the admin page to check the order record.
1.4. At the order detail page, it should show `Payment Method: Omise` as usual.

screen shot 2560-01-04 at 2 20 08 am

2. Because `AbstractOmiseClient::getModuleVersion()` was also updated, we should test if plugin can retrieve the module version and pass to the Omise API properly.
The step are as follows:
2.1. Go to Omise Dashboard, check a charge transaction that we just created.
2.2. Check request log, at `User Agent`, it should show `OmiseMagento/2.0.0` as usual.

screen shot 2560-01-04 at 2 15 56 am

#### 4. Impact of the change

No.

#### 5. Priority of change

Normal

#### 6. Additional Notes

- To run unit test, please check http://devdocs.magento.com/guides/v2.0/test/unit/unit_test_execution_cli.html


guzzilar:
Updated a related ticket, `T2553`

guzzilar:
👍

guzzilar:
Note, removed `Status: Ready` tag, since approach on PR #38 was declined.

PR #44, opened 2017-01-24
Support JPY, IDR and SGD currencies

guzzilar:
> This PR required PR #40 to be merged first.

#### 1. Objective

To make plugin support the following currencies.
- JPY (Japan)
- IDR (Indonesia)
- SGD (Singapore)

**Related information**:
Related issue(s): 🙅
Related ticket(s): T2244, T2245

#### 2. Description of change

- Add condition, if store's currency is not one of "THB", "JPY", "IDR", "SGD". Omise payment method will show warning message that the current currency is not support with Omise service.
 
- Be able to create charge with "THB", "JPY", "IDR", "SGD" currencies.
- Remove `Currency` configuration from the payment setting page, now using store's base currency configuration instead.
- Format the display of transaction amount at the plugin dashboard page.
- Remove placeholder that mentioned about `satang` from `create transfer input field` at the plugin dashboard page.
- Remove condition to hide `create transfer input field` if balance is lower than the minimum-transfer-amount. (this form used to support only `THB`, now I removed it out and let API says error if merchant cannot create transfer).

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.1
- **Omise plugin version**: Omise-Magento 1.9.0.6
- **PHP version**: 5.6.28

**✏️ Details:**

1. ✅ If enter to the payment setting page (`System > Configuration > Sales > Payment Method`). "Currency" configuration is gone.
  screen shot 2560-01-24 at 2 56 08 pm

2. ✅ Test checkout with "JPY" currency
    2.1 Go to `System > Configuration > General > Currency Setup` to change base currency to `Japanese Yen`

    2.2. Set product's price (in this test, I set it to `20,000 YEN`) and do normal checkout

    screen shot 2560-01-24 at 5 27 32 pm

    2.3. Check result from Omise log, `currency` must be `JPY` and amount must be the same as the product price (0 decimal, no-subunit) (i.e. set product price to `15,025`, the amount that send to Omise must be `15025`)

    screen shot 2560-01-24 at 5 32 03 pm

3. ✅ Test checkout with "IDR" currency
    3.1. Go to `System > Configuration > General > Currency Setup` to change base currency to `Indonesian Rupiah`

    3.2. Set product's price (in this test, I set it to `Rp20,000`) and do normal checkout

    screen shot 2560-01-24 at 5 06 05 pm

[...]

PR #42, opened 2017-01-20
Update Omise-PHP library to version 2.6.0

guzzilar:
> This PR required PR #40 to be merged first. 
>
> Note, 👆  just because I rebased PR #40 into this branch before submit pull request. Technically, 2 PRs are not related to each other at all.

#### 1. Objective

Update Omise-PHP to the latest version, 2.6.0.

For more information about Omise-PHP 2.6.0, please check the link below.
https://github.com/omise/omise-php/releases/tag/v2.6.0

**Related information**:
Related issue(s): 🙅
Related ticket(s): 🙅

#### 2. Description of change

- Update Omise-PHP.
- Remove unused directories & files, the list are as follows:
  1 `./tests`
  2 `.gitignore`
  3 `circle.yml`
  4 `composer.json`
  5 `phpunit.xml`

#### 3. Quality assurance

**🔧 Environments:**

- **Platform**: Magento 1.9.3.1
- **Omise plugin**: Omise-Magento 1.9.0.6
- **PHP**: 5.6.28
- **Omise-PHP**: 2.6.0

**✏️ Details:**

To prove that this changed will not effect to the current function, we should consider 2 parts,
✅ 1. Test if plugin still able to connect with Omise API, go to `Plugin Dashboard` page.
    screen shot 2560-01-20 at 4 19 36 pm

✅ 2. Test create charge with normal checkout process, API should retrieve new `Omise-PHP` version, which is, 2.6.0.
    👇 I created new order, and this is the charge log.
    screen shot 2560-01-20 at 4 23 22 pm

#### 4. Impact of the change

The steps below should be mentioned in README.md in a new section `how to upgrade Omise-Magento 1.9.0.6 to 1.x`

To completely remove some unused files in `/lib/omise-php`
1. At the Magento's root directory, remove folder `./lib/omise-php`
2. Then, copy new `./lib/omise-php` from this changes and paste into `./lib`

Says, if you're using Macbook and want to copy `lib/omise-php` folder from plugin to Magento directory, click `Replace`. (not `Merge`)
screen shot 2560-01-20 at 4 36 25 pm

_**note**, this action isn't must-do action, even user doesn't follow this step, it's totally fine._
_Since this step is just to remove the following files_
```
- Remove unused directories & files, the list are as follows:
  1 `./tests`
  2 `.gitignore`
  3 `circle.yml`
  4 `composer.json`
  5 `phpunit.xml`

```

#### 5. Priority of change

Normal.

#### 6. Additional Notes

[...]

PR #46, opened 2017-01-31
Introduce payment strategy classes to handle payment action.

guzzilar:
> ⚠️ This PR required PR #44 to be merged first.

#### 1. Objective

Refactor the PaymentMethod class for the extensibility, polymorphism and reduce code complexity that usually, will keep maintain in only one file (Omise_Gateway_Model_PaymentMethod).

- Before:
  All of logics have to be in `Model/PaymentMethod.php` since this class is the main class that Magento will import and execute the commands (authorize, capture, refund, etc).

- After:
  All of logics were moved to each of sub-classes (`Model/Strategies/*Strategy.php`).
  So, `Model/PaymentMethod.php` will just only trigger a proper `payment strategy` class instead (not handle logic/validation by itself).

This approach would also be helpful for `3-D Secure` & `Internet Banking` methods, even for another features (i.e. Refund). By introducing `Strategy` classes. (for more detail, please check `6. Additional Notes`).

_Note, to create new payment method 'Internet Banking', Magento doesn't allow to have 2 payment method at the same class, we have to separate it between `CreditCard` and `InternetBanking`_

screen shot 2560-02-06 at 1 21 04 pm

**Related information**:
Related issue(s): 🙅
Related ticket(s): 🙅

#### 2. Description of change

- Introduce `Omise_Gateway_Model_Payment` (`Model/Payment.php`) as a main-base class for `payment method`.
- Introduce payment strategy approach, `Model/Strategies/StrategyInterface` & `Model/Strategies/StrategyAbstract`
- Introduce concrete strategy classes, `AuthorizeStrategy` and `CaptureStrategy`, `ManualCaptureStrategy`.
- Update `Model/PaymentMethod.php` corresponding to the changes, use the strategy approaches to make a charge/capture.

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.1.
- **Omise plugin version**: Omise-Magento 1.9.0.6
- **Omise PHP version**: 2.4.0
- **PHP version**: 5.6.28

**✏️ Details:**

1. ✅ Test checkout an order with `authorize only` action.
    _**expectation**: success as normal, charge will be created, order will be created_

    1.1. At plugin setting page, set `payment action = authorize only`.

    1.2. Do normal checkout process, it must success (redirect to success page).

    1.3. Check at Omise dashboard, you will see a new charge transaction in the charge list, with pending status.
    screen shot 2560-02-01 at 6 17 17 am

    1.4. All of request parameters should be the same as before.
    screen shot 2560-02-01 at 6 17 41 am

2. ✅ Test checkout an order with `authorize only` action and use [failed card test](https://www.omise.co/api-testing#failed-charge)
    _**expectation**: charge will be failed and show error message on a screen as normal_

    2.1. Do normal checkout process, but test checkout with [failed card test].

    2.2. Error will be shown in a popup box.
    screen shot 2560-02-01 at 12 24 54 pm

    _note, this tests were tested with cards, `Invalid Account Number`, `Payment Rejected`, `Insufficient Fund`. See [failed card test document](https://www.omise.co/api-testing#failed-charge)_

3. ✅ Test checkout with `authorize and capture` action
    _**expectation**: success as normal, charge will be created, order will be created_

    3.1. At plugin setting page, set `payment action = authorize and capture`.

    3.2. Do normal checkout process, it must success (redirect to success page).

    3.3. Check at Omise dashboard, you will see a new charge transaction in the charge list, with successful status.
    screen shot 2560-02-01 at 1 14 20 pm

    3.4. All of request parameters should be the same as before.
    screen shot 2560-02-01 at 1 14 40 pm

4. ✅ Test checkout an order with `authorize and capture` action and use [failed card test]
    4.1. Same as second test but error message is slightly different,
    screen shot 2560-02-01 at 1 22 54 pm

5. ✅ Test manual capture
    5.1. Go to admin, order list page. (`Sales > Orders`)

    5.2. Choose one order and click `Invoice` to capture a transaction.
    screen shot 2560-02-01 at 6 19 57 am copy

    5.3. Check Omise dashboard, charge log, the dashboard shows, `capture (manual)` with `requested capture` log.
    screen shot 2560-02-01 at 1 31 41 pm

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

**Known issues:**
- Note, https://github.com/omise/omise-magento/pull/46/files#diff-252896e00d37e3f160a971e0827a3ef2R35
    Ideally, this class shouldn't take care about how to set `public_key` and `secret_key` (and shouldn't contains `protected $public_key` & `protected $secret_key` by itself).
    These should be provided at `Configuration` class.

- `Omise_Gateway_Model_Omise` will be deprecated in the future.
    For payment gateway methods, use `Omise_Gateway_Model_Payment` instead.

- This PR is preparing for `3-D Secure` & `Internet Banking`.
    `Model/Strategies/OffsiteInternetBankingStrategy` will be provided for Internet Banking feature in another PR.

- `Omise_Gateway_Model_PaymentMethod` will be renamed to `Omise_Gateway_Model_CreditCardPayment` in the future, to separate between `Credit Card` payment behaviour and `Internet Banking` payment behaviour.

- Continue from above, then we can have `Omise_Gateway_Model_OffsiteInternetBankingPayment` to handle only `Internet Banking` payment behaviour.

### DIFF
--- a/src/app/code/community/Omise/Gateway/Model/Payment.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment.php
@@ -60,7 +60,7 @@ abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abs
         $this->defineOmiseKeys();
 
         try {
-            $result = $strategy->process($this, $params);
+            $result = $strategy->perform($this, $params);
         } catch (Exception $e) {
             Mage::throwException(Mage::helper('payment')->__($e->getMessage()));
         }
--- a/src/app/code/community/Omise/Gateway/Model/Payment.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment.php
@@ -58,6 +58,8 @@ abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abs
     public function perform(Omise_Gateway_Model_Strategies_StrategyInterface $strategy, $params)
     {
         $this->defineOmiseKeys();
+        $this->defineOmiseApiVersion();
+        $this->defineOmiseUserAgent();
 
         try {
             $result = $strategy->perform($this, $params);
@@ -86,6 +88,28 @@ abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abs
 
         if (! defined('OMISE_SECRET_KEY')) {
             define('OMISE_SECRET_KEY', $secret_key ? $secret_key : $this->secret_key);
+        }
+    }
+
+    /**
+     * @param  string $version
+     *
+     * @return void
+     */
+    protected function defineOmiseApiVersion($version = '2014-07-27')
+    {
+        if (! defined('OMISE_API_VERSION')) {
+            define('OMISE_API_VERSION', $version);
+        }
+    }
+
+    /**
+     * @return void
+     */
+    protected function defineOmiseUserAgent()
+    {
+        if (! defined('OMISE_USER_AGENT_SUFFIX')) {
+            define('OMISE_USER_AGENT_SUFFIX', 'OmiseMagento/1.9.0.6 Magento/' . Mage::getVersion());
         }
     }
 }
--- a/src/app/code/community/Omise/Gateway/Model/Payment.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment.php
@@ -21,6 +21,11 @@ abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abs
     protected $secret_key;
 
     /**
+     * @var \Mage_Sales_Model_Order_Payment
+     */
+    protected $payment_information;
+
+    /**
      * Load necessary file and setup Omise keys
      *
      * @return void
@@ -50,19 +55,33 @@ abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abs
     }
 
     /**
+     * @return \Mage_Sales_Model_Order_Payment
+     */
+    public function getPaymentInformation()
+    {
+        return $this->payment_information;
+    }
+
+    /**
      * @param  \Omise_Gateway_Model_Strategies_StrategyInterface $strategy
-     * @param  mixed                                             $payment
+     * @param  \Mage_Sales_Model_Order_Payment                   $payment
+     * @param  int|float                                         $amount
      *
      * @return Mage_Payment_Model_Abstract
      */
-    public function perform(Omise_Gateway_Model_Strategies_StrategyInterface $strategy, $params)
-    {
+    public function perform(
+        Omise_Gateway_Model_Strategies_StrategyInterface $strategy,
+        Mage_Sales_Model_Order_Payment $payment,
+        $amount
+    ) {
         $this->defineOmiseKeys();
         $this->defineOmiseApiVersion();
         $this->defineOmiseUserAgent();
 
+        $this->payment_information = $payment;
+
         try {
-            $result = $strategy->perform($this, $params);
+            $result = $strategy->perform($this, $amount);
         } catch (Exception $e) {
             Mage::throwException(Mage::helper('payment')->__($e->getMessage()));
         }
--- a/src/app/code/community/Omise/Gateway/Model/Payment.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment.php
@@ -1,0 +1,91 @@
+<?php
+abstract class Omise_Gateway_Model_Payment extends Mage_Payment_Model_Method_Abstract
+{
+    /**
+     * @var \Omise_Gateway_Model_Config
+     */
+    protected $config;
+
+    /**
+     * Omise's public key
+     *
+     * @var string
+     */
+    protected $public_key;
+
+    /**
+     * Omise's secret key
+     *
+     * @var string
+     */
+    protected $secret_key;
+
+    /**
+     * Load necessary file and setup Omise keys
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        // Load Omise-PHP library
+        require_once(Mage::getBaseDir('lib') . '/omise-php/lib/Omise.php');
+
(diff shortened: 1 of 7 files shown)


guzzilar:
rebased with PR #42 `1-update-omisephp-lib`

guzzilar:
rebased with PR #44 `1-dev-currency-support`

guzzilar:
rebased with PR #44

guzzilar:
rebased with `1-stable`

nimid:
@guzzilar 

The implementation of strategy pattern is not proper. Including apply it with the current Magento 1 core source code, it is not fit.

1. To implement this strategy design pattern, it requires and produces too many and unnecessary duplicate code.
2. With increase of strategy, it raise the number of class (strategy) that requires maintenance.
3. Some strategies are not totally different, for example:
    - `Omise_Gateway_Model_Strategies_AuthorizeStrategy`
    - `Omise_Gateway_Model_Strategies_CaptureStrategy`
They are only one different request parameter, `capture`. They are not totally different. So, they do not need to be entirely separated.
4. This design is different from the Omise API. For example: Omise charge

| Payment Method  | Omise API | Strategy Pattern |
| --- | --- | --- |
| Normal credit card | `Charge.create()` | `Authorize.perform()` |
| 3-D Secure | `Charge.create(return_uri)` | `AuthorizeThreeDSecure.perform()` |
| Internet banking | `Charge.create(return_uri, offsite)` | `OffsiteInternetBanking.perform()` |

For Omise API, each method has been separated by using the parameter but the strategy pattern each method become the class.

This pull request is the introducing of strategy pattern. Therefore, the problem has not been clearly expressed. The problems will be clear at the implementation, the latest of the series of pull requests that finish the features. For examples, the pull requests #53 and #57.

**In order to emphasize the duplicate code**
A tool, [PHP Copy/Paste Detector (PHPCPD)](https://github.com/sebastianbergmann/phpcpd), can be easily used to emphasize this technical debt, the duplicate code.

For example to run the command:
```
phpcpd --min-tokens=30 --exclude=omise-php/ src/
```

The table below shows the result after run the above command on 2 branches.

| Branch  | Duplicate Code (%) | Total Lines of Code | Files |
| --- | --- | --- | --- |
| [`1-stable`](https://github.com/omise/omise-magento/tree/1-stable) (before change) | 0.52 | 954 | 1 clones with 5 duplicated lines in 1 files |
| [Merged `1-dev-3ds-validate-callback` and `1-dev-ib-handle-callback`](https://github.com/nimid/omise-magento) (latest change) | 12.36 | 2055 | 9 clones with 254 duplicated lines in 8 files |

**Note:**
The tool, PHP Copy/Paste Detector, has been used to quickly and automatically count the duplicate code. A parameter `--min-tokens` can be adjusted. Even though, using the default value of a parameter `--min-tokens` which is 70, the duplicate code is still found.

**In order to emphasize the implementation of strategy pattern that it is not fit with Magento 1 core code**
For the strategy pattern, it require the same interface for every strategy but at least in [this line](https://github.com/omise/omise-magento/pull/53/files#diff-56b13928e59bd4e6b04edbd04d96e9d7R52), it requires another condition to be a layer before pass to process.

The technical debt that cannot be avoided or can be managed or has been planned to pay its interest is acceptable.

To simplify or reduce the complexity of source code, it has other solutions that can be complied, conformed, followed or similar to the OmiseAPI or Magento 1 core code and has lower impact in part of technical debt, which it is more appropriate.

guzzilar:
@nimid 
> 1. To implement this strategy design pattern, it requires and produces too many and unnecessary duplicate code.

> 2. With increase of strategy, it raise the number of class (strategy) that requires maintenance.

> 3. Some strategies are not totally different, for example

That's my intention.

The benefit of `single responsibility` principle is to separate class and 1 class perform as less as possible.

So, if we break the code, core code or the rest classes will not break because we maintain each action per class. More easy to maintain and refactor or whatever since 1 fix not effect to all of the rest.
And that's why `interface` class comes.

Think about if we need to implement `refund` action.
We don't need to be worried for the rest of the code. We just create new `RefundStrategy` class. and then, as from `interface` says, that strategy class need to implement `perform()` and `validate()` methods.

So, we can make sure that all of the implementation, even from unknown developer will implement the same way as us and not break the core function.


> 4. This design is different from the Omise API. For example: Omise charge

Note, Omise API didn't tell how you have to implement. I don't see the point of this part.

guzzilar:
Correct typo.
```
The benefit of Single Responsibility principle  (to separate class and 1 class perform as less as possible),

So, if we break the code, core code or/and the rest classes will not break because we maintain each action per class. More easy to maintain and refactor or whatever since 1 fix not effect to all of the rest.
And that's why interface class comes.
```

guzzilar:
@nimid 
> This pull request is the introducing of strategy pattern. Therefore, the problem has not been clearly expressed. The problems will be clear at the implementation, the latest of the series of pull requests that finish the features. For examples, the pull requests #53 and #57.

Not quite understand this comment.
Sure, because this was just refactoring code to prepare for coming PRs.
PR doesn't meant to just only solve a problem but even refactoring or just correct typo/ update document and as objective says:

> Refactor the PaymentMethod class for the extensibility, polymorphism and reduce code complexity that usually, will keep maintain in only one file (Omise_Gateway_Model_PaymentMethod).


> For the strategy pattern, it require the same interface for every strategy but at least in this line, it requires another condition to be a layer before pass to process.

Yup, because I can't find a proper solution to handle 3-D Secure payment from Magento core code.


> To simplify or reduce the complexity of source code, it has other solutions that can be complied, conformed, followed or similar to the OmiseAPI or Magento 1 core code and has lower impact in part of technical debt, which it is more appropriate.

Please suggest :)

guzzilar:
Note, for me, maintain 20 of classes is not bad. The bad design is maintain 2000 lines with 20 methods in 1 class.

guzzilar:
👍

PR #50, opened 2017-02-07
Display a message, if the current currency is not supported

nimid:


guzzilar:
@nimid 

another solution is, have another template file and in that file just says only 
```html
<?php $_code = $this->getMethodCode(); ?>

<ul id="payment_form_<?php echo $_code; ?>" class="form-list" style="display:none;">
    <li>
        <div class="validation-advice">
            Checkout with Credit Card (Powered by Omise) is not support for the current currency.
        </div>
    </li>
</ul>
```

and move that condition into `Omise_Gateway_Block_Form_Cc` class.
by have another method, something like, `isApplicable()`.
i.e.
```php
<?php
class Omise_Gateway_Block_Form_Cc extends Mage_Payment_Block_Form
{
    {
        parent::_construct();
        $this->isApplicable();
    }

    public function isApplicable()
    {
        if (! $this->isCurrentCurrencySupported()) {
            $this->setTemplate('payment/form/inapplicable-method.phtml');
        }

        $this->setTemplate('payment/form/omisecc.phtml');
    }

    public function isCurrentCurrencySupported()
    {
        // ...
    }

    // ...
}
```

That we can get rid of that conditional block from html file.
And also, we can keep template file to do its function without thinking how to display the content.

guzzilar:
merged here https://github.com/omise/omise-magento/pull/50 as discussed at https://github.com/omise/omise-magento/pull/44#issuecomment-278048436

I added tag `status: invalid` and going to close it.

PR #41, opened 2017-01-20
[1.x series] Introduce 3-D Secure payment feature 

guzzilar:
> This PR required PR #46 to be merged first. 

#### 1. Objective

To support 3-D Secure feature.
Merchant will be able to enable 3-D Secure config and process a 3-D Secure payment through the plugin.

> This changes support for payment "Authorize only" & "Authorize and Capture" actions.

**Related information**:
Related issue(s): #18 
Related ticket(s): 🙅

#### 2. Description of change

Task list:

- [x] Add `3-D Secure` option to the setting page (see PR #43).
- [x] Redirect buyer to the 3-D Secure page if the setting is enabled (see PR #47).
- [x] Implement `callback` url/function to validate charge (see PR #53).
- [x] Update Magento's order status (see PR #53).

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.1.
- **Omise plugin version**: Omise-Magento 1.9.0.6.
- **PHP version**: 5.6.14.

**✏️ Details:**

Please check at all sub-tasks (PR #43, #47, #53).

#### 4. Impact of the change

Please check at all sub-tasks (PR #43, #47, #53).

#### 5. Priority of change

Normal

#### 6. Additional Notes

Nothing.

PR #52, opened 2017-02-09
Separate between class that taking care of configuration and payment action.

guzzilar:


guzzilar:
Was implemented at #41 PR.

PR #60, opened 2017-03-08
Bump release to v1.10

guzzilar:
Bump release to v1.10.

- Update README.md
-Update CHANGELOG.md
- Update version number from v1.8 to v1.10

> **TL;DR;**
> To make our plugin be able to install through the [Magento Connect Store](https://www.magentocommerce.com/magento-connect/omise-1.html) Im going to skip `v1.9` and tag with `v1.10` instead.
>
> ...
>
> Because **Magento extension management** can't do upgrade or downgrade an extension that a new version number is lower than the current one, which is, Omise-Magento's version was `v1.9.0.6` before we changed our versioning structure.
> So, if we go with `v1.9`. Merchant will not be able to upgrade our plugin through Magento extension management (Magento Connect).
> 
> At this page 👇 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+### [v1.10 _(March 9, 2017)_](https://github.com/omise/omise-magento/releases/tag/v1.10)
+
+#### ✨ Highlights
+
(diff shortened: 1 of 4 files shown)


guzzilar:
👍

PR #66, opened 2017-03-31
Redirect buyer out to the internet banking page if the placeOrder action is succeeded

guzzilar:
#### 1. Objective

To handle the placeOrder action, either redirect buyer out to the internet banking page (if the action is succeeded) or raise an error message onto the screen (if the action is failed).

**Related information**:
Related issue(s): 🙅

#### 2. Description of change

**_• Implement RESTFul API `GET orders/:order_id/omise-offset`_**

> Because Magento 2 moved the `placeOrder` action out to the client side, and place an order through Ajax > RESTful API. But unfortunately, that API's response can't send a payment method detail or  `authorize_uri` back, so we have to make new API and make a request again to retrieve that information.

**_• Update JS model to make a redirection and handle either success/fail case._**

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

1) ✅ Once buyer place an order with internet banking payment, buyer will be redirected to the internet banking page.
    1) Just do normal checkout process and submit (`place order`).
    2) Buyer will be redirected out to the internet banking page.
    screen shot 2560-03-31 at 5 58 00 pm

2) ✅ Test access an API directly, user would be rejected. (because user's session is not match with the user's cart).
    screen shot 2560-03-31 at 6 50 18 pm

3) ✅ In case something wrong with a new API, an error message will be raised as the below  screenshot.
    screen shot 2560-03-31 at 6 54 43 pm

#### 4. Impact of the change

Now buyer will be redirected out to the internet banking page after place an order.

#### 5. Priority of change

Normal.

#### 6. Additional Notes

- This PR doesn't include the callback part, so, once buyer finish the internet banking process, buyer will be redirect back to a store, and then, store will raise an `404 page not found` error.

### DIFF
--- a/Api/Data/PaymentInterface.php
+++ b/Api/Data/PaymentInterface.php
@@ -1,0 +1,51 @@
+<?php
+namespace Omise\Payment\Api\Data;
+
+interface PaymentInterface
+{
+    // Fields
+    const TYPE          = 'object';
+    const ORDER_ID      = 'order_id';
+    const AUTHORIZE_URI = 'authorize_uri';
+
+    /**
+     * @api
+     *
+     * @param  int $value
+     *
+     * @return self
+     */
+    public function setOrderId($value);
+
+    /**
+     * @api
+     *
+     * @param  string $value
+     *
+     * @return self
+     */
+    public function setAuthorizeUri($value);
+
+    /**
+     * Always return a "payment" string.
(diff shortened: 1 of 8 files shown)


guzzilar:
👍

PR #75, opened 2017-04-10
Prepare the code structure for 3-D Secure payment.

guzzilar:
#### 1. Objective

To prepare core code structure to support an implementation of the 3-D Secure payment.

**Related information**:
Related issue(s): 🙅

#### 2. Description of change

**_• Integrate payment with an `initialize` gateway command, to use it as a middle layer to determine which final command that it will proceed with._**
The reason why we need this, is because in Magento system, they've only provided 2 commands as a basic/ general credit card payment method use case.

For example, for the `Authorize and Capture` command. The general Magento flow will be 
```
1. Place Order
    > 2. Try submit payment
        > 3. if payment success (captured), create an invoice
            > 4. update status to "processing"
```

But as for 3-D Secure payment behavior, we need to hold the result at the step 3.
So, the flow is gonna be something like below.
```
1. Place Order
    > 2. Try submit payment
        > 3. if payment success (just successs on create a transaction), update status to "pending payment"

4. Trigger callback
    > 5. Validate the charge result
        > 6. If success, create an invoice
            > 7. Update order status to `processing`
```
👆 To go with this flow, we need to reimplement the payment flow by ourself by using `initialize` gateway command.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16

**✏️ Details:**

_The main purpose of these tests is to make sure that this change will not effect to the current charge functions_
 
1. ✅ Test create charge with `authorize only` payment action.

2. ✅ Test create charge with `authorize and capture` payment action.

3. ✅ Test create charge with any payment action, and make it fail by using **[Failed Card Test](https://www.omise.co/api-testing#failed-charge)**

#### 4. Impact of the change

You must clear Magento caches to get this new payment method.

1. At Magento admin page, go to `SYSTEM > Cache Management`
2. Click `Flush Magento Cache`. (or you can choose to disable all caches while testing).
    screen shot 2560-03-22 at 5 32 08 pm

#### 5. Priority of change

Normal

#### 6. Additional Notes

1. The implementation of 3-D Secure payment haven't been implemented yet. This PR just prepare the core structure to support the coming PRs.

### DIFF
--- a/Gateway/Command/CreditCardStrategyCommand.php
+++ b/Gateway/Command/CreditCardStrategyCommand.php
@@ -1,0 +1,110 @@
+<?php
+namespace Omise\Payment\Gateway\Command;
+
+use Magento\Payment\Gateway\Command\CommandException;
+use Magento\Payment\Gateway\Command\CommandPoolInterface;
+use Magento\Payment\Gateway\CommandInterface;
+use Magento\Payment\Gateway\Helper\ContextHelper;
+use Magento\Payment\Gateway\Helper\SubjectReader;
+use Magento\Sales\Model\Order;
+use Omise\Payment\Model\Config\Cc as Config;
+
+class CreditCardStrategyCommand implements CommandInterface
+{
+    /**
+     * @var string
+     */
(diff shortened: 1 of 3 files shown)


guzzilar:
👍

PR #80, opened 2017-04-11
Refactor & correct the callback functions behavior

guzzilar:
#### 1. Objective

To reduce some duplicated code and correct the callback behaviors on both Internet Banking & 3-D Secure.

**Related information**:
Related issue(s): 🙅

#### 2. Description of change

1. Reduce some duplicated conditional block in a charge validation of Internet Banking callback
    _Use `Validator\CaptureResultValidator` instead._

1. Correct the order cancellation behavior

1. Won't `restore a quote` even payment is failed.
    _An issue on Magento caused wrong amount of quantity in the product catalog. (if execute `registerCancellation` and `restoreQuote` together, an actual quantity amount on a product catalog will be doubled)._

1. Update Gateway\Validation's `validateResponse()` method visibility back to `protected`.

1. Correct DocBlock.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

1. ✅ Test checkout with some amount of product with internet banking payment, then, make a charge to be failed, and check the catalog, a product quantity should be released back with the same amount that user has ordered.

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/Controller/Callback/Internetbanking.php
+++ b/Controller/Callback/Internetbanking.php
@@ -163,6 +163,18 @@ class Internetbanking extends Action
      * @param \Magento\Sales\Model\Order       $order
      * @param \Magento\Framework\Phrase|string $message
      */
+    protected function invalid(Order $order, $message)
+    {
+        $order->addStatusHistoryComment($message);
+        $order->save();
+
+        $this->messageManager->addErrorMessage($message);
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


guzzilar:
👍

PR #82, opened 2017-04-11
Allow Omise none 3-D Secure account to process with the same behavior as 3-D Secure payment

guzzilar:
#### 1. Objective

To prevent 1 use case that, if Omise enables 3-D Secure payment to a merchant account but merchant doesn't set `3-D Secure option` to `Yes` at the plugin side, it will breaks all of charge transactions.

This changes allow none 3-D Secure account to be able to process with the same behavior as 3-D Secure account. (plugin behavior).

**Related information**:
Related issue(s): 🙅

#### 2. Description of change

1. Remove conditional block that check if charge.status is `pending` out from the validation

2. At the `ThreeDSecureCommandResponseValidator` class, recheck at the last step to make sure if that charge transaction is really failed or it's just normal charge from none 3-D Secure account.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

1. ✅ With Omise none 3-D Secure account, set `3-D Secure` option at the Magento payment setting page to `yes`. Then, try to make a charge as normal (both `authorize` and `authorize and capture`). It should work as normal.

#### 4. Impact of the change

No.

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/Gateway/Response/PendingInvoiceHandler.php
+++ b/Gateway/Response/PendingInvoiceHandler.php
@@ -11,21 +11,12 @@ class PendingInvoiceHandler implements HandlerInterface
      */
     public function handle(array $handlingSubject, array $response)
     {
-        /** @var bool **/
-        $captured = $response['data']['captured'] ? $response['data']['captured'] : $response['data']['paid'];
+        /** @var \Magento\Payment\Gateway\Data\PaymentDataObjectInterface **/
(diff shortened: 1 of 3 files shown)


guzzilar:
👍

PR #86, opened 2017-05-22
Provides a solution for One Step Checkout extensions.

guzzilar:
#### 1. Objective

One Step Checkout makes difference behavior of a checkout process by put every input fields that buyer need to fill into 1 checkout page, then use JavaScript-AJAX to load, process, validate all of fields in that checkout page.

By that, usually that Omise-Magento relies on a `CONTINUE` button at the **Payment Information** step in order to trigger **Omise.js** to create an Omise token when buyer click the button.
screen shot 2017-05-22 at 15 01 03

Which is, unsupported on One Step Checkout extension's behavior (OSC has only 1 button to 'Place Order' an order).
Means, Omise token isn't created properly > No Omise Token passing through Magento server side = Omise raise an error `Need Omise's key.`.

So, this PR provides another solution to support OSC behavior by auto trigger **Omise.js** when Omise's credit card fields are filled to create an Omise token.

**Related information**:
Related issue(s): 🙅 

#### 2. Description of change

1. Provide new setting, `osc_support`.
    

2. Update JavaScript at the `payment/form/omisecc.phtml` to support the OSC behavior by binding `onBlur` event, then check if card form is ready to be sent to create Omise token or not when buyer move 'focus' out from a field (onBlur event is triggered).

3. Update `Need Omise's key` error message to another message to make it more understandable.
    screen shot 2017-05-23 at 17 01 13

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.2.2, 1.9.3.2.
- **PHP version**: 5.6.30

**✏️ Details:**

There are 2 main cases that we need to consider.

1. Test make a charge with normal checkout form with/without enable OSC Support feature.
2. Test make a charge with One Step Checkout extension with/without enable OSC Support feature.

**1. Test make a charge with normal checkout form with/without enable OSC Support feature.**

Without installing One Step Checkout extension, you can test fill the card form and observe whether card form can create a token properly or not.

1.1. ✅ Disable _OSC_Support_ feature, and make a charge as normal.

1.2. ✅ Enable _OSC_Support_ feature, and make a charge.

Here, you will see that card form does difference behavior.

Without **OSC Support** feature, card form will create a token when buyer click a 'CONTINUE' button on a page. But with **OSC Support**, Omise.js will be triggered in order to create a token after buyer fills all card information fields and leave focus from the field.

**Note**, With normal checkout form, disabled _OSC Support_ is recommended.

**2. Test make a charge with One Step Checkout extension with/without enable OSC Support feature.**
_Here, I was testing with [LotusBreath](https://www.magentocommerce.com/magento-connect/lotusbreath-one-step-checkout.html) extension_

2.1. ✅ Disable _OSC_Support_ feature. It will raise an error as below (note: Previously, Omise will return `Need Omise's key`). The reason, please check the section `1. Objective`.
    screen shot 2017-05-23 at 17 01 13

2.2. ✅ Enable _OSC_Support_ feature, and make a charge.

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

**Note**  
This solution might not work with all of One Step Checkout extensions.
For any who is using any of One Step Checkout extension but can't make a charge with this solution, feel free to submit your query through our support team at support@omise.co or create new ticket at https://github.com/omise/omise-magento/issues

### DIFF
--- a/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
+++ b/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
@@ -64,6 +64,16 @@ class Omise_Gateway_Block_Form_Cc extends Mage_Payment_Block_Form
     }
 
     /**
+     * Whether the One Step Checkout Support option is enabled
+     *
+     * @return bool
+     */
+    public function isOscSupportEnabled()
+    {
+        return Mage::getModel('omise_gateway/paymentMethod')->isOscSupportEnabled();
+    }
+
+    /**
      * Retrieve Omise keys from database
      *
      * @return string|array
--- a/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
+++ b/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
@@ -13,7 +13,7 @@ class Omise_Gateway_Block_Form_Cc extends Mage_Payment_Block_Form
     {
         if ($this->isApplicable()) {
             $this->setTemplate('payment/form/omisecc.phtml');
-        } else {
+         } else {
             $this->setTemplate('payment/form/omise-inapplicable-method.phtml');
         }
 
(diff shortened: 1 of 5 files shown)


guzzilar:
👍

PR #102, opened 2017-12-12
MAG1 : Add multi-currency support.

guzzilar:
#### 1. Objective

There are lots of requests asking for make Omise-Magento plugin to support multi-currency feature since the day we've released this feature.

So here we are! 😃 

#### 2. Description of change

- **Only show `Internet Banking` and `Alipay` payment methods when using 'THB' currency in a store**
  Note that Internet Banking and Alipay payments can't be operated with other currencies and only able to be used with an Omise account that belongs to Thailand account.

- **Using `BaseCurrencyCode` instead of `getOrderCurrencyCode`.**
  The different between `BaseCurrencyCode` and `getOrderCurrencyCode` is that, in Magento currency system, you can set **base currency** and **display currency** of your store.

  Base currency will be the one that Magento uses to pass around 3rd party payment gateways and display currency just, to display on a screen purpose.

  So after researched on Magento's multi currency behaviours and checked at Magento default payments' implementation. It seems that Magento tends to do currency conversion at the Magento side before passing to a payment gateway to process a payment, and just use order's currency and amount for display purpose only.

  That's why we changed from `$order->getOrderCurrencyCode` to `$order->getBaseCurrencyCode`.

- **Display Omise Credit Card form when using the following currencies in a Magento store**
  JPY, THB, SGD, USD, EUR, GBP.

#### 3. Quality assurance

**🔧 Environments:**

- **Magento**: Open Source 1.9.3.6.
- **PHP**: 5.6.30.

**✏️ Details:**

_[drafting]_

#### 4. Impact of the change

None.

#### 5. Priority of change

Normal

#### 6. Additional Notes

- Document: https://www.omise.co/multi-currency

- Note that only Omise Thailand Account that can use `multi-currency` feature as Omise mentioned in the above document:
  ```
  Multi-currency is currently in beta testing, and is only available to merchants
  with a Thai-registered Omise account. We'll be adding support for other countries
  in the coming months. 
  ```

### DIFF
--- a/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
+++ b/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
@@ -49,18 +49,10 @@ class Omise_Gateway_Block_Form_Cc extends Mage_Payment_Block_Form
      */
     protected function isStoreCurrencySupported()
     {
-        $currencyCode = Mage::app()->getStore()->getBaseCurrencyCode();
-
-        switch ($currencyCode) {
-            case 'THB':
-            case 'JPY':
-            case 'IDR':
-            case 'SGD':
-                return true;
-                break;
-        }
-
-        return false;
+        return in_array(
+            Mage::app()->getStore()->getBaseCurrencyCode(),
+            array('JPY', 'THB', 'SGD', 'IDR', 'USD', 'EUR', 'GBP')
+        );
     }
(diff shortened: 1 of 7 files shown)


guzzilar:
rebased with `1-stable` branch

guzzilar:
👍

PR #103, opened 2017-12-20
MAG2: Introduce API Object model structure and refactoring code

guzzilar:
#### 1. Objective

This PR is aiming to pay the technical debt that I was creating when I first initiated this extension.
By that I mean to refactoring code by giving a proper organize in the core code and also, to introduce API object classes.

#### 2. Description of change

• **Clean up code responsibilities**
For example, previously `Gateway/Http/Client/Payment.php` had been taking care too much of responsibilities such as, needed to handle on how to communicate with Omise-PHP library, needed to retrieve module version, needed to know how to create charge and so on.

Now those responsibilities have been assigned to a proper classes and let them do as less as it can.

• **Introduce Model\Api classes**
Well, some conditions have been rewritten again and again especially for checking a charge result.
This change introduces a better way to handle those duplicated code by having a `Model\Api\Object` and let them do their jobs so we can avoid 'human-error' from duplicated code as much as we can.

• **Refactoring validator classes**
From the second point, here is to show how we can make the code way better from the above benefit.

#### 3. Quality assurance

**🔧 Environments:**

- **Magento Open Source**:  `v2.2.1`.
- **PHP**: `v7.0.25`.

**✏️ Details:**

Make sure that all payment methods can work properly by create a new charge with 'Credit Card', 'Credit Card 3-D Secure', 'Internet Banking' and observe in both success and fail cases.

#### 4. Impact of the change

None.

#### 5. Priority of change

Normal

#### 6. Additional Notes

None.

### DIFF
--- a/Gateway/Http/Client/Payment.php
+++ b/Gateway/Http/Client/Payment.php
@@ -2,12 +2,11 @@
 namespace Omise\Payment\Gateway\Http\Client;
 
 use Exception;
-use Magento\Framework\App\ProductMetadataInterface;
-use Magento\Framework\Module\ModuleListInterface;
 use Magento\Payment\Gateway\Http\ClientInterface;
 use Magento\Payment\Gateway\Http\TransferInterface;
 
 use Omise\Payment\Model\Config\Config;
+use Omise\Payment\Model\Omise;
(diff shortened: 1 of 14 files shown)


guzzilar:
👍

Issue #84, opened 2017-04-15
[ReflectiontionException]


prasertkv:
-Magento v2.1.6
When use command  "setup:di:compile"  Will result below
![image](https://cloud.githubusercontent.com/assets/27570520/25065315/0c1f2292-2238-11e7-978a-fdceb7ec3a04.png)

guzzilar:
Hi @prasertkv 

Can you try remove file at `app/code/Omise/Payment/Model/Ui/OmiseConfigProvider.php` then, run the command again?

Please let me know if this problem still occurs.
Nam.

guzzilar:
Hi, this issue has been solved at Omise-Magento v2.3.
ref: https://github.com/omise/omise-magento/releases/tag/v2.3

PR #106, opened 2018-01-30
MAG2: Removed unused files

guzzilar:
### 1. Objective

There is an error `Class Omise\Payment\Model\Config` does not exist when execute `setup:di:compile` command which occurred because some unused files are still referring to that class (that `Omise\Payment\Model\Config` class has been removed since a while but there are still some files that refer to that class).

**Related information**:
Related issue(s): #84 

#### 2. Description of change

Just removed unused files so those will not be included when `setup:di:compile` is executed

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento Open Source v2.2.2
- **PHP version**: 7.0.25.

**✏️ Details:**

From the issue #84, if you try execute `setup:di:compile` command this time, it won't show any errors.
screen shot 2561-01-30 at 9 24 26 pm

#### 4. Impact of the change

None.

#### 5. Priority of change

Normal

#### 6. Additional Notes

None.


guzzilar:
👍

PR #107, opened 2018-01-30
MAG2: Introduce Webhook feature

guzzilar:
### 1. Objective

There are some cases that we need to consider using WebHooks feature.

- **An order that has been placed at Magento store with `payment_action=authorize only`. Then later, merchant performs `capture` action at Omise dashboard instead.**

    Without Webhook, Magento store will never knew that that charge transaction has been _captured_ already at Omise dashboard.
    Means, we are going to need something like a callback feature to tell Magento store that that `captured` event is triggered. So Magento store can update an order status according to the result. 

- **Some offsite payments don't resolve users' payment at a time they (user) complete their payment process. For example, Internet Banking, sometimes Bank cannot response the payment result back to Omise API at a time user 'submit' their payment.**

    User will be redirected back to Magento store with `charge.status = pending` causing that an order status won't be changed to `processing` nor `failed` until merchant manual change it.

    Again, we are going to need something like a callback feature to tell Magento store that that `charge transaction` is resolved. So Magento store can update an order status either to make it as 'canceled' or 'completed'. 

**Related information**:
Ref: https://www.omise.co/api-webhooks

### 2. Description of change

1. New field, `Webhook endpoint` at the payment setting page.

2. Add a new route `/omise/callback/webhook` to handle Omise Webhook's callback.

3. Add capability to handle Omise Webhook: `charge.complete` event.
    Once you enable the Webhook feature, Omise-Magento will be able to receive Omise Webhook's payload from the `charge.complete` event and proper update the order status according to the charge result.

### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento Open Source v2.2.2
- **PHP version**: 7.0.25.

**✏️ Details:**

1. **Make sure that a `payment-review` order gets update from the Omise Webhook: `charge.complete` event properly.**
    > Note: only **3-D Secure payment** and **Internet Banking Paymentv methods that can trigger the `charge.complete` event.

    1.1. For 3-D Secure payment case, the `Model\Event` handler class should do nothing. The `charge.complete` event should effect only a `payment-review` order (which shouldn't happen with 3-D Secure payment)

    1.2. For Internet Banking: successful case.
        Make sure that once Webhook's request arrives at the store, the order status, invoice, transaction will be updated to a proper status.

    1.3. For Internet Banking: failed case.
        Make sure that once Webhook's request arrives at the store, the order status, invoice, transaction will be updated to a proper status.

2. Once Omise Webhook is triggered and made a request back to the Magento store. Make sure that the event handler class won't be trigger if an order status isn't set to `payment-review`.
    In some case, especially Internet Banking payment: the payment result may be resolved at the time user is redirected back to the store, causing that the order status may be set to a proper status before the request from Webhook arrive.

    Means that we have to make sure that once the order status won't get update twice from above case.

#### 4. Impact of the change

None.

### 5. Priority of change

Normal

### 6. Additional Notes

- There are some events that this PR hasn't implemented yet (i.e. `charge.create`, `charge.capture`, `charge.reverse`, `refund.create`).
    For more information, please check https://www.omise.co/api-webhooks.

### DIFF
--- a/Block/Adminhtml/System/Config/Form/Field/Webhook.php
+++ b/Block/Adminhtml/System/Config/Form/Field/Webhook.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Omise\Payment\Block\Adminhtml\System\Config\Form\Field;
+
+class Webhook extends \Magento\Config\Block\System\Config\Form\Field
+{
+    /**
+     * @var \Magento\Framework\Url
+     */
+    protected $urlHelper;
+
+    /**
+     * @param \Magento\Backend\Block\Template\Context $context
+     * @param \Magento\Framework\Url                  $urlHelper
+     * @param array                                   $data
+     */
+    public function __construct(
+        \Magento\Backend\Block\Template\Context $context,
+        \Magento\Framework\Url $urlHelper,
+        array $data = []
+    ) {
+        $this->urlHelper = $urlHelper;
+        parent::__construct($context, $data);
+    }
+
+    protected function _getElementHtml(\Magento\Framework\Data\Form\Element\AbstractElement $element)
+    {
+        return $this->urlHelper->getRouteUrl('omise/callback/webhook', ['_secure' => true]);
+    }
+}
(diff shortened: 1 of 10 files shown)


guzzilar:
👍

PR #108, opened 2018-03-01
1 handle quote submit failure event

guzzilar:
### 1. Objective

There are some edge-cases that causing the order-submit process cannot be completed (either from an unexpected issue from Magento or any incident that interrupt the order-submit flow).

In that case, Magento will revert everything back to the state before the order is submitted (cancel invoice, return an item back to the inventory, remove order record from the DB, etc).

Problem is, this incident could be happened even after payment is processed, causing that, at Omise side, buyer's card will be charged as normal but there is no record at the Magento side.

So we have to do refund or reverse those charge as well in that case.

--

**Note**: Thanks Gogoprint team for the report.

### 2. Description of change

- Implement and register an observer class to handle `sales_model_service_quote_submit_failure` event.

- Integrate with Omise Refund API and Omise Charge Reverse API.

### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.7.
- **PHP version**: 5.6.30

**✏️ Details:**

Edge case is hard to reproduce, however, there is a way to do hard-code in the Magento core file to make it 'throws' an error before complete the order-submit process so we can test how Magento handles this incident.

1. Go to `app/code/core/Mage/Sales/Model/Order.php` then, find `place()` method and edit:
```php
/**
 * Place order
 *
 * @return Mage_Sales_Model_Order
 */
public function place()
{
    Mage::dispatchEvent('sales_order_place_before', array('order'=>$this));
    // throw new Exception(); // Add this line to throw an error before place the payment.
    $this->_placePayment();
    // throw new Exception(); // Add this line to throw an error after place the payment.
    Mage::dispatchEvent('sales_order_place_after', array('order'=>$this));
    return $this;
}
```

2. Select order and do checkout as usual. At the last step of checkout process, once you click `Place Order`, Magento will raise an error that `there was an error processing your order.` (error message comes from Magento)
    screen shot 2561-03-01 at 16 59 00

3. If you throw an error before place the payment: Nothing happen. (no request to Omise, no record on the Magento order list page).

4. If you throw an error after place the payment, there will be 2 scenarios
    4.1 If the `payment action` option is set to `authorize only`.
        Charge transaction will be reversed.
        screen shot 2561-03-01 at 17 06 55
    4.2 If the `payment action` option is set to `authorize and capture`.
        Charge transaction will be voided.
        screen shot 2561-03-01 at 17 04 25

### 4. Impact of the change

None

### 5. Priority of change

Normal

### 6. Additional Notes

This issue will not effect to the Internet Banking and Alipay payment methods.
Anyway, if the incident happens with Internet Baking or Alipay payment methods, those charges will just stay as `pending` and get `expired` later on.


guzzilar:
👍

PR #88, opened 2017-08-18
Added credit card refund capability

jonrandy:
#### 1. Objective

Add online refund functionality for credit card payments

#### 2. Description of change

Added refund functions to `Omise_Gateway_Model_PaymentMethod` along with a new `RefundStrategy` - following the pattern of previously added Omise functionality. Also added a simple function for converting amounts to subunit amounts based upon the currency code.

#### 3. Quality assurance

Tested on a local Magento 1 installation against an Omise account in test mode. No testing done in 'Live' mode as yet

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.4
- **Omise plugin version**: Omise-Magento 1.12 (?)
- **PHP version**: 5.6.31

**✏️ Details:**

Create and pay for an order using Omise credit card payment method, go to that order's invoice, click 'credit memo', locate the refund section, click 'refund online'.

#### 4. Impact of the change

The only visible impact of the change is that the 'refund online' button will now appear on the order's invoice screen in the admin area

#### 5. Priority of change

Normal

#### 6. Additional Notes

My first PR!

PR #109, opened 2018-03-01
Add credit card refund capability.

guzzilar:
> Note: this PR is a re-opened PR developed under the current implementation (the one that has removed the strategy pattern).
>
> The original PR and all solution inspirations came from PR #88 (all credit goes to @jonrandy 😄)

### 1. Objective

It's hard to go back n forth between Magento Admin page and Omise Dashboard to refund a charge transaction.

For example, if you want to refund a specific charge, you have to
- Search for a specific order on Magento Admin
- Copy charge id from Magento Admin, then go back to Omise Dashboard
- Search for a specific charge transaction by using the charge id from Magento Admin
- Make refund
- Then go back to Magento Admin, to update order status (create a creditmemo)
- And so on.

So this PR offers an easier way to create a refund right in the Magento Admin page so you don't need to go out from your store anymore!

### 2. Description of change

- Add credit card refund capability, so now merchant can create a refund right in the Magento Admin page without jumping out to Omise Dashboard.

### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 1.9.3.7
- **PHP version**: 5.6.31

**✏️ Details:**

1. Create and pay for an order using Omise credit card payment method, then go to that order's invoice page, click 'credit memo', locate the refund section, click 'refund online'.
  refund-001
  refund-002

2. Once done, you will see that there is a `refunded` record in the order detail page.
  refund-003
  refund-004

3. And, here is the result when you check at the Omise Dashboard page.
  refund-005

4. In case if something happens and the refund process can't be completed, an error will be raised at the Magento Admin page as below:
  screen shot 2561-03-02 at 16 36 52

### 4. Impact of the change

The only visible impact of the change is that the 'refund online' button will now appear on the order's invoice screen in the admin area

### 5. Priority of change

Normal

### 6. Additional Notes

@jonrandy's first PR!

### DIFF
--- a/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
@@ -24,7 +24,9 @@ class Omise_Gateway_Model_Payment_Creditcard extends Omise_Gateway_Model_Payment
     protected $_isGateway        = true;
     protected $_canAuthorize     = true;
     protected $_canCapture       = true;
+    protected $_canRefund        = true;
     protected $_canReviewPayment = true;
+
 
     /**
      * flag if we need to run payment initialize while order place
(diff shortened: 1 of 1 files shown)


guzzilar:
Updated description.

guzzilar:
👍

PR #117, opened 2018-05-16
Switches on 'showInStore' for all config options to allow full multi-store functionality

jonrandy:
#### 1. Objective

Make sure all config settings for Omise Magento are available down to the store level - to facilitate full multistore capability. Also add some details about the current store in the charge metadata

#### 2. Description of change

`adminhtml/system.xml` modified to turn on all `showInStore` switches.
`paymentdatabuilder.php` modified to add store info to charge metadata

#### 3. Quality assurance

I set up a sample site with multiple stores and views, and different Omise Magento settings for each store view. I then made an order from each store view, checking that the payment options were correct for each view, and that the correct store info had been added to the charge metadata in each case

#### 4. Impact of the change

No visible changes will be seen when this update is installed unless installed on a site that has multiple stores/views - in which case the merchant will have control over the plugin on a per store basis

image

👆 When you select any 'scope' after the update, the Omise options will be available for changing. They were only available at the top level before

#### 5. Priority of change

Normal

#### 6. Additional Notes

It's not raining today, but wow is it hot! 🔥

### DIFF
--- a/Gateway/Request/PaymentDataBuilder.php
+++ b/Gateway/Request/PaymentDataBuilder.php
@@ -45,12 +45,19 @@ class PaymentDataBuilder implements BuilderInterface
         $payment = SubjectReader::readPayment($buildSubject);
         $order   = $payment->getOrder();
 
+        $store_id = $order->getStoreId();
+        $om = \Magento\Framework\App\ObjectManager::getInstance();
+        $manager = $om->get('Magento\Store\Model\StoreManagerInterface');
+        $store_name = $manager->getStore($store_id)->getName();
+
         return [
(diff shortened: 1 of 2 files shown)


guzzilar:
@jonrandy can you provide a screenshot of, 'what the setting page looks like' with and without these change? 
Also would be nice to have a screenshot of the change in the charge metadata.

jonrandy:
@guzzilar Did I do that right?

guzzilar:
@jonrandy Uhmm.. probably no. Just `git rebase master` is enough actually.

jonrandy:
I thought that's what I did

PR #118, opened 2018-05-21
Removes link/dependency with Authorize.net module

jonrandy:
#### 1. Objective

Remove a small link/dependency with the `authorize.net` module. It's risky leaving it there as we cannot guarantee the module will be present. This makes our module more self contained

#### 2. Description of change

Add a new class within our module to handle the `PaymentAction` dropdown for credit cards in the admin area (copied from the similar class in Authorize.net)

#### 3. Quality assurance

Go to the magento admin area and check the dropdown mentioned above is still functioning correctly

#### 4. Impact of the change

No noticeable change for users

#### 5. Priority of change

Normal

#### 6. Additional Notes

What causes thunder anyway?

### DIFF
--- a/Model/Source/PaymentAction.php
+++ b/Model/Source/PaymentAction.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Omise\Payment\Model\Source;
+
+use Magento\Framework\Option\ArrayInterface;
+
+class PaymentAction implements ArrayInterface
+{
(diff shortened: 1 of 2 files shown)


guzzilar:
> What causes thunder anyway?

These guys 👇 
screen shot 2561-06-04 at 11 39 56

PR #120, opened 2018-05-24
Updates to some config as recommended by Magento docs

jonrandy:
I was reading through the official docs on Magento for developing payment gateway modules, and I noticed we have not done some of the required config. Not entirely sure exactly HOW necessary these things are, but probably best to do things 'by the book'

### DIFF
--- a/etc/module.xml
+++ b/etc/module.xml
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Omise_Payment" setup_version="2.3.0" />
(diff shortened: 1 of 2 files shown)


etc/module.xml
```diff
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Omise_Payment" setup_version="2.3.0" />
+    <module name="Omise_Payment" setup_version="2.3.0">
+    	<sequence>
+    		<module name="Magento_Sales"/>
+    		<module name="Magento_Payment"/>
+    		<module name="Magento_Checkout"/>
```
guzzilar: @jonrandy just tab-to-space thing here 😄

jonrandy: image

guzzilar: I know 😜

PR #127, opened 2018-06-25
restrict alipay for thb transactions only

jacekstan:
#### 1. Objective

Alipay is available only for Thai Baht orders.
The purpose of this PR is to disable possibility of make Alipay payments for orders in other currencies.

#### 2. Description of change

Changed in JavaScript rendering files and HTML Alipay Checkout payment view.

To resolve issue there is checking if order is done in THB (Thai Baht) if yes, than Alipay payment method is displayed normally, otherwise option is _readonly_ and it is displayed message 
_Option available only for Thai Baht currency orders_

screen shot 2018-06-25 at 22 16 08

#### 3. Quality assurance
**🔧 Environments:**

- **Platform version**: Magento CE 2.4.
- **PHP version**: 7.0.29.

**✏️ Details:**

Step 1. Open magento admin, make sure that Thai Baht is main store currency.
Make order, and do payment with Alipay.
Step 2. Open magento admin, change Thai Baht store currency to other currency,
Try to make payment with Alipay. Payment should not be available.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A


guzzilar:
@jacekstan can you try one more case, a case where store currency is `THB` but order currency is `USD`?

jacekstan:
I added checking if store currency is THB.

view/frontend/web/template/payment/offsite-alipay-form.html
```diff
@@ -8,9 +8,13 @@
                value: getCode(),
                checked: isChecked,
                click: selectPaymentMethod,
-               visible: isRadioButtonVisible()" />
+               visible: isRadioButtonVisible(),
+               enable: isActive()" />
         <label data-bind="attr: {'for': getCode()}" class="label">
             <span data-bind="text: getTitle()"></span>
+            <span 
+                data-bind="i18n: '(option available only for orders in Thai baht)',
```
guzzilar: @jacekstan Uhmm I'm still 50/50 on this.
Because in fact, it doesn't matter what your order's currency is. The store's currency has to be the same.

i.e. if store currency is set to `usd` and buyer choose to order in `thb`, it still won't work with this condition.

Maybe we just really need to hide the payment method if store currency isn't THB?

jacekstan: I agree but I also think that would be good question to PM ?

jacekstan: @guzzilar as agreed with Big, we are keeping option, but disabled.

PR #133, opened 2018-07-05
2017-11-02 api support

jacekstan:
#### 1. Objective

Migration plugin to use new API 2017-11-02 to support new payment methods like: Bill Payment Tesco Lotus Solution

#### 2. Description of change

Used no create source approach, described in last sections of both documents named: *Creating a charge by adding attribute `source[type]`*.
https://www.omise.co/alipay
https://www.omise.co/offsite-payment

Most changes were done in request construction.
Renamed parameter used to send information about type of payment.
In old api it was called `offsite` in new api it is called `source[type]`

Charge is created in single api call.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.
- **Omise API Version**: 2017-11-02

**✏️ Details:**

To properly test it make successful payments using Internet Banking and Alipay method.

#### 4. Impact of the change

This change force all users using plugin `omise-magento` to use `API 2017-11-02`.
This change does not require from users to switch version of plugin in Omise Dashboard and is necessary for to make available new payment methods.

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/Gateway/Request/PaymentOffsiteBuilder.php
+++ b/Gateway/Request/PaymentOffsiteBuilder.php
@@ -14,7 +14,12 @@ class PaymentOffsiteBuilder implements BuilderInterface
     /**
      * @var string
      */
-    const OFFSITE = 'offsite';
+    const SOURCE = 'source';
+
+    /**
+     * @var string
+     */
+    const SOURCE_TYPE = 'type';
 
     /**
      * @var string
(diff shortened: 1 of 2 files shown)


Gateway/Request/PaymentOffsiteBuilder.php
```diff
@@ -49,10 +49,10 @@ public function build(array $buildSubject)
         
         switch ($method->getMethod()) {
             case Alipay::CODE:
-                $paymentInfo[self::OFFSITE] = 'alipay';
+                $paymentInfo[self::SOURCE_TYPE] = 'alipay';
```
guzzilar: @jacekstan this one a bit strange that we have `source[type]` as a string while in fact, `type` should be assigned in an array.. how about we assign this one as array instead? 
Something like
```php
    'type' => 'alipay'
]
```

or 
```php
    self::SOURCE_TYPE => 'alipay'
]
```

PR #137, opened 2018-07-19
code refactoring, joined offsite request classes to one renamed to Om…

jacekstan:
#### 1. Objective

This pull request is purely to decrease code and make it more readable. 

**Related information**:
#### 2. Description of change

Offsite Payments like Alipay and Internet Banking are using exactly the same Request Builder Virtual Type in `di.xml`. Tesco Lotus will also use exact one.
Instead of making for each Requrest builder seperated for each Offsite Payment Method it is good to have the same one.

so we had following classes: 
- `OmiseOffsiteInternetbankingRequest`
- `OmiseOffsiteAlipayRequest`
- and now `OmiseOffsiteTescoRequest`

with the same content, now it will be one class called:
`OmiseOffsiteRequest`

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.

**✏️ Details:**

To properly test it:
- download changes
- clean cache using command: `bin/magento cache:flush`
- make payments with InternetBanking and Alipay

#### 4. Impact of the change

N/A

#### 5. Priority of change

High.

#### 6. Additional Notes

N/A

### DIFF
--- a/etc/di.xml
+++ b/etc/di.xml
@@ -48,20 +48,11 @@
 
     <virtualType name="OmiseOffsiteInternetbankingInitializeCommand" type="Magento\Payment\Gateway\Command\GatewayCommand">
         <arguments>
-            <argument name="requestBuilder" xsi:type="object">OmiseOffsiteInternetbankingRequest</argument>
+            <argument name="requestBuilder" xsi:type="object">OmiseOffsiteRequest</argument>
             <argument name="transferFactory" xsi:type="object">Omise\Payment\Gateway\Http\TransferFactory</argument>
             <argument name="client" xsi:type="object">OmiseCharge</argument>
             <argument name="handler" xsi:type="object">OmiseOffsiteResponseHandler</argument>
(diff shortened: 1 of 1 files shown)


etc/di.xml
```diff
@@ -110,50 +101,56 @@
...
         <arguments>
             <argument name="handlers" xsi:type="array">
                 <item name="paymentDetails" xsi:type="string">Omise\Payment\Gateway\Response\PaymentDetailsHandler</item>
+                <item name="pendingPayment" xsi:type="string">Omise\Payment\Gateway\Response\PendingPaymentHandler</item>
+                <item name="pendingInvoice" xsi:type="string">Omise\Payment\Gateway\Response\PendingInvoiceHandler</item>
             </argument>
         </arguments>
     </virtualType>
 
-    <virtualType name="OmiseOffsiteResponseHandler" type="Magento\Payment\Gateway\Response\HandlerChain">
```
jacekstan: this part of the code was moved bit below, and grouped with Offsite Hanlders, that is why it shows falsely that class was renamed.

Issue #104, opened 2018-01-10
Cancelling payment with Internet Banking


SzamanSZ:
In Magento 1.x, using the newest plugin (multicurrency), when I choose Internet Banking, cancel even on bank login page and go back to Merchant, I am redirected to success page without paying.

guzzilar:
@SzamanSZ Hi, thanks for your report, I can also reproduce this issue.

SzamanSZ:
Any approximation about when this issue will be fixed?

jacekstan:
closing as bug is fixed.

PR #142, opened 2018-08-29
Cancel Internet Banking (and other offsite payments) fix

jonrandy:
#### 1. Objective

Adds a message to the user informing them that the payment process is still ongoing if they get to the success screen when the charge status is stilll `pending`

(Also moves some template files to an `omise` folder to match the pattern found elsewhere)

**Related information**:
Related issue(s): #104 

#### 2. Description of change

Added a single line of code to display a notice for the user in the situation described above. Moved some template files to an `omise` folder

#### 3. Quality assurance

Engineered the `pending` charge situation through temporary hardcoding, tested to make sure the message appeared to the user.

- **Platform version**: Magento CE 1.9.3.9
- **PHP version**: 5.6.23

#### 4. Impact of the change

screen shot 2018-08-28 at 12 05 30

#### 5. Priority of change

Normal

#### 6. Additional Notes

Probably my smallest PR ever?


guzzilar:
@jonrandy Just wonder, how this commit https://github.com/omise/omise-magento/pull/142/commits/34e293888d0b2aa6c7f9e33beb3e11569d1e2df9 related to the fix? 
Can the issue be fixed without this one?

jonrandy:
It was stuff I needed to do to get a dev environment and working version of
Magento 1 and the plugin up and running. The function definition change was
causing a PHP error, and moving the templates made sense as all other
places we have templates they were in an `omise` subfolder. Doing this also
eases the creation of a 'virtual install' of the plugin in a docker
environment. If I hadn't done this, these changes would have needed to be
in a PR of their own.

guzzilar:
I mean, I know it is needed to be done but if it is not related to the fix then shall we have a new PR for this commit?

SzamanSZ:
Do Thai banks have no other way of **really** fixing it?

Now I've just checked with the v. 1.15 and the issue is still there. I chose Internet Banking as the payment method, SCB bank and I cancelled logging into the bank. Then after going "Back to merchant", I was redirected to the success page with your added message (but I cancelled the login! Not even the payment). Now imagine I'm a real client and just decided to change the payment method, use different bank account or whatever. Instead of being redirected back to the cart/checkout page, I am redirected to the success page, and if I want to buy the product, I must go through the entire purchasing process again (find the right product, add it to the cart, etc.).

Not a satisfying solution at all.

PR #144, opened 2018-08-29
Removes specific internet banking fee messages and replaces them with a generic one

jonrandy:
#### 1. Objective

Remove the specific fee information labels for internet banking since these fees are subject to change quite frequently (in fact they are already out of date). Replace them with a generic message telling the user that the bank may charge them a small fee for the transaction.

#### 2. Description of change

The HTML for the internet banking block was altered as described above

#### 3. Quality assurance

Checked to see it looked okay

#### 4. Impact of the change

User now always receives appropriate information about charges, regardless if the banks change the charge amounts. Maintenance burden for this for us is also removed.

screen shot 2018-08-29 at 17 00 51

#### 5. Priority of change

Normal

#### 6. Additional Notes

💵 💵 💵


src/app/design/frontend/base/default/template/payment/form/omise/omiseoffsiteinternetbankingpayment.phtml
```diff
@@ -56,8 +57,11 @@ $code = $this->getMethodCode();
             </div>
             <div class="omise-banking-text-wrapper">
                 <span class="title">Bangkok Bank</span><br/>
-                <span class="rate secondary-text">Fee: 10 THB (same zone), 20 THB (out zone)</span>
+                <!-- <span class="rate secondary-text">Fee: 10 THB (same zone), 20 THB (out zone)</span> -->
```
jacekstan: why we need that commented?
it is what github is for..to remember that stuff ;)
I think we could simply delete that commented lines

src/app/design/frontend/base/default/template/payment/form/omise/omiseoffsiteinternetbankingpayment.phtml
```diff
@@ -56,8 +57,11 @@ $code = $this->getMethodCode();
...
             </div>
         </label>
     </li>
+    <li>
+        <span class="rate secondary-text"><?php echo($feeMessage); ?></span>
```
jacekstan: why we just don't echo that:
$this->__('The bank may charge a small fee for the transaction.');
instead of setting it to variable used only once.

or do better: 
`<span class="rate secondary-text"><?= $this->__('The bank may charge a small fee for the transaction.'); ?></span>`
btw. is that ok.. for php5? `<?= ?>`

src/app/code/community/Omise/Gateway/Block/Form/Offsiteinternetbankingpayment.php
```diff
@@ -11,7 +11,7 @@ class Omise_Gateway_Block_Form_Offsiteinternetbankingpayment extends Mage_Paymen
     protected function _construct()
     {
         parent::_construct();
-        $this->setTemplate('payment/form/omiseoffsiteinternetbankingpayment.phtml');
+        $this->setTemplate('payment/form/omise/omiseoffsiteinternetbankingpayment.phtml');
```
jacekstan: is that part of another PR? that code?
you probably did not checkout from `master` this branch?

src/app/design/frontend/base/default/template/payment/form/omise/omiseoffsiteinternetbankingpayment.phtml
jonrandy: I forgot to delete 😝

src/app/code/community/Omise/Gateway/Block/Form/Offsiteinternetbankingpayment.php
jonrandy: I did, but another PR is incoming (waiting approval) that has that. My local project will not work without it so I needed it here too. It will all come out in the wash

src/app/design/frontend/base/default/template/payment/form/omise/omiseoffsiteinternetbankingpayment.phtml
jonrandy: I was trying to keep any PHP code as separate as possible, but thinking about it, it's probably better to have the text actually where it appears in the HTML - better for editing

src/app/code/community/Omise/Gateway/Block/Form/Offsiteinternetbankingpayment.php
jacekstan: After bit thinking if it is since 5.4 I think we should add than we should keep:
`<?php echo "text" ?>`
Because we have no idea.. what kind of servers are used by customers.
And since it is Magento 1, I think better to play it safe.
And additionally I have checked magento 1.9.2.4 and there is no single place where they use construction `<?= ?>`

jonrandy:
https://secure.php.net/manual/en/migration54.new-features.php
screen shot 2018-08-30 at 10 46 21

PR #146, opened 2018-08-30
Offsite Refactoring

jacekstan:
Tesco lotus uses the same functionality, and tesco is not offsite method, so we need to change names for more general.

we are not sure how will look future api calls for new payment methods,
Now Offsite functionality is common also for Offline payments like Tesco Lotus,
that is why in this PR my proposal is to have 2 general descriptions.
OmiseXXXXXX
and OmiseCCXXXXX

I can't find better name for offline/offsite functionalities, so I put it to general ones.
and just simply removed _offsite_ word from common functionalities.

#### 1. Objective

This PR is necessary as for offline _tesco lotus_ functionality to use the same parts of code as offsite payments (alipay internet banking).

#### 2. Description of change

Just names of classes,

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.

**✏️ Details:**

To properly test it you need to make payments with offsite Alipay, Internet Banking, and with Credit Card method.

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -9,8 +9,8 @@ use Magento\Sales\Model\Order;
 use Magento\Sales\Model\Order\Payment\Transaction;
 use Omise\Payment\Model\Omise;
 use Omise\Payment\Model\Api\Charge;
-use Omise\Payment\Model\Config\Offsite\Internetbanking;
-use Omise\Payment\Model\Config\Offsite\Alipay;
+use Omise\Payment\Model\Config\Internetbanking;
+use Omise\Payment\Model\Config\Alipay;
 
 class Offsite extends Action
 {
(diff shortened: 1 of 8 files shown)


guzzilar:
Uhmm.. I feel a bit strange about this.. like, this change is trying to remove `offsite` and `offline` concept and make it into a generic name where those 2 methods can share its functionality.
But the name is too generic and it's not fit with every payment methods.

The class `Payment*` stuff does name like a base for everything but in fact it is only made for `offline` and `offsite` methods. Which make it more ambiguous to understand..

Just like we are having a `base` class but the class is not truly a base class cuz it made for specific usage.

Hmmm let me think a bit.. will try to come up to propose you with some solution..

PR #148, opened 2018-09-06
Instalment Payment Option Implementation

jacstn:
Instalment Payment Integration.

#### 1. Objective
Introduce Installment Payment Option in Magento 2 plugin.

#### 2. Description of change
Added support for Installment Payments.

Key notes:
- minimum order is hardcoded in `JavaScript` payment rendering file as 5000 THB, if minimum order is not met than plugin behave in the same way as it is in situation when user choose different payment currency than THB
- payment option is only available in THB orders
- translations has been added to Thai Language only.
- payment cancellation on backend side (after forwarding to bank site) will be implemented once `Omise Backend` will support sending cancellation event.
- currently after payment is cancelled by user after redirection to bank page it is treated as `pending` payment and cart is cleared by `Magento`, in that case user need to re-do shopping.
- renamed styles file. File `offsite-internetbanking.css` had definitions of styles used in checkout internet banking payment method. Renamed file to general `style.css` and extended styles for supporting displaying installment payment method.
- renamed `CCConfigProvider.php` file/class to more generic `OmiseConfigProvider.php` as included also `omise capabilities` object.

screenshot 2018-12-18 at 15 26 05

Display message when 5000 THB value of order is not met.

image

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.

**✏️ Details:**

To properly test it make successful payment with Installment Option and Credit Card Payment Method.

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

This PR requires https://github.com/omise/omise-php/pull/99 to be merged first.
This PR requires new version of `omise-php` library to be released with above `Capabilities` PR


Model/Capabilities.php
```diff
@@ -0,0 +1,34 @@
+<?php
+namespace Omise\Payment\Model;
+
+use Omise\Payment\Model\Api\Capabilities as OmiseCapabilitiesAPI;
+use Omise\Payment\Model\Omise;
+
+class Capabilities
```
jonrandy: Again, should this really be called `capabilities`?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -0,0 +1,178 @@
...
+                };
+            },
+
+            /**
+             * Is method available to display
+             *
+             * @return {boolean}
+             */
+            isActive: function () {
+                return this.getOrderCurrency().toLowerCase() === 'thb' && this.getStoreCurrency().toLowerCase() === 'thb';
```
jonrandy: Seems very wrong to be hardcoding this given that the `Capabilities` API gives us the currencies that are applicable for each backend, but I guess it works for now and helps to get Instalment payments `out of the door` faster

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -0,0 +1,178 @@
...
+            },
+
+            /**
+             * Returns Installment Terms
+             *
+             * @return {array}
+             */
+            getInstallmentTerms(id) {
+                const capabilities = checkoutConfig.payment.capabilities;
```
jonrandy: Again, this should really be named 'Backends'? To match conventions in OmisePHP and Capabilities API

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -0,0 +1,178 @@
...
+                return window.checkoutConfig.quoteData.store_currency_code;
+            },
+
+            /**
+             * Check if order value meets minimum requirement
+             *
+             * @return {boolean}
+             */
+            orderValueTooLow: function () {
+                return window.checkoutConfig.totalsData.base_grand_total < 5000;
```
jonrandy: This should be coming from capabilities API, or at least comment why it is hardcoded for now

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+            <div role="alert" class="messages">
+                <div class="message-warning warning message">
+                    <div data-bind="i18n: 'Option available only for orders in THB'"></div>
+                </div>
+            </div>
+        </div>
+        <div data-bind="visible: orderValueTooLow()" class="page messages">
+            <div role="alert" class="messages">
+                <div class="message-warning warning message">
+                    <div data-bind="i18n: 'Minimum order value is 5000 THB'"></div>
```
jonrandy: Shouldn't really be hardcoding this, it will come from the Capabilities API response

Model/Api/Capabilities.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Omise\Payment\Model\Api;
+
+use OmiseCapabilities;
+
+class Capabilities extends Object
```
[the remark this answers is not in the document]
jacstn: Idea of that functionality is getting capabilities of user.
the api is also called capabilities in omise-php
call is also to backend about capabilities.

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+               click: selectPaymentMethod,
+               visible: isRadioButtonVisible(),
+               enable: isActive() && !orderValueTooLow()" />
+        <label data-bind="attr: {'for': getCode()}" class="label">
+            <span data-bind="text: getTitle()"></span>
+        </label>
+        <div data-bind="visible: !isActive()" class="page messages">
+            <div role="alert" class="messages">
+                <div class="message-warning warning message">
+                    <div data-bind="i18n: 'Option available only for orders in THB'"></div>
```
jonrandy: Surely, the option should not even display for THB orders? So this message would not be needed?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: what even mean backends? :) it is about getting capabilities, user capabilities, what he can do.. from logical point of view.. and naming convention it is fine IMO

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: It will be subject for future PR

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+                            </div>
+                            <div class="omise-banking-text-wrapper">
+                                <span class="title" data-bind="i18n: 'Krungthai Card'"></span>
+                            </div>
+                        </label>
+                        <select style="margin-top:10px" data-bind="options: getInstallmentTerms('ktc'),
+                                        optionsText: function(item) {
+                                            return item;
+                                        },
+                                        valueAllowUnset: true,
```
jonrandy: Seems like it just displays a list of numbers? We should really be telling the user somewhere that this is the number of monthly payments. A number alone probably is not enough

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: it is consistent with what we have now.. it is visible with other payment methods, but not clickable. HTML control it is disabled.

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+        <div class="omise-banking-fee-warn" data-bind="i18n: 'Choose your installment terms'"></div>
+        <form class="form" data-bind="attr: {
+              id: getCode() + 'Form',
+              }">
+
+            <fieldset data-bind="attr: {class: 'fieldset payment items ' + getCode(), id: 'payment_form_' + getCode()}">
+                <ul class="form-list">
+                    <!-- KTC -->
+                    <li class="items">
+                        <input type="radio" name="offsite" value="installment_ktc" data-bind="
```
jonrandy: Why are these harcoded? i.e. each individual instalment type? They should be coming from the Capabilities API response (dynamically generated)

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: it will be subject for future PR

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: Good idea... this will be subject for future PR, for now we have only limited number specified providers.

Model/Capabilities.php
jacstn: I think so.. it is about getting capabilities.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: this is hardcoded in exactly the same way for other payment methods, but good idea... will change it in future PR.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: I described it in PR but maybe could add it into code..

jacstn:
I have commented, Improvements with automation of fulfilling forms will be done later when Capabilities api will be more mature, and actually will be really corresponding with what you can do. Like you have notice that 5000 thb is not even working with backend now.

Model/Api/Capabilities.php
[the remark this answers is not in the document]
jacstn: but capabilities will develop, and when that will get shape I might do refactoring. For now I like to leave it as Capabilities

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: The capabilities API returns a range of things that the merchant account is capable of. The backends part of this lists the payment backends that the account is capable of using. Other parts detail the limits dictating the amounts a merchant account is capable of charging and transferring. In the API response, each capability within the capabilities returned in the `capabilities` object is contained in it's own, named, section. In the case of a lot of the code here, we are interested in just the 'backend' capabilities of the account, not all the capabilities together. For this reason, and to avoid any confusion between the various levels (API, OmisePHP, Plugin), I feel it is best we stick to using the same names as suggested/defined in the API response

jonrandy: I imagine more capabilities (in their own sections) will be added into the API response in the future also

Model/Api/Capabilities.php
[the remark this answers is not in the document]
guzzilar: I was intending to have `Model/Api/*` structure to be represented as a db-layer (but fetch from API instead of real DB). Capabilities sounds good to me. 

But the way to execute this class is a bit ambiguous.. 
`Model\Api\Capabilities::read()` > `Model\Capabilities::get()` > then OmiseConfigProvider got `Installment's backend`.
Would be nice to refactor a bit to explicitly shows its intention. (also, I think this layer should not do filter by itself. The filter one should be in `Model\Capabilities` and keep this one simple and does only retrieving data from API job)

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: Second that. `capabilities` is too vague.

Model/Api/Capabilities.php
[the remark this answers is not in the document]
jonrandy: @guzzilar Re-read the code - the object here is getting the `backends` info from within capabilties object, but it calls itself `capbilities`. This creates confusion we could do without

Model/Ui/OmiseConfigProvider.php
```diff
@@ -5,8 +5,10 @@
 use Magento\Payment\Model\CcConfig as MagentoCcConfig;
 use Omise\Payment\Model\Config\Cc as OmiseCcConfig;
 use Omise\Payment\Model\Customer;
+use Omise\Payment\Model\Capabilities;
```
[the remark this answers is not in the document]
jonrandy: Whether or not the option to switch instalment payments on/off is available in admin should surely be dependent upon the existence of instalment backends within the account's `capabilities` object?

Model/Api/Capabilities.php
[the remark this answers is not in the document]
guzzilar: @jonrandy yeah I got that. I mean, this could instead be
```php
// Model\Api\Capabilities
class Capabilities
{
    public function retrieve()
    {
        // ..
    }
}
```

And just that. Just a layer to connect with Omise API to fetch data back to its object. 
Then `Model\Capabilities` can provide a method to filtering backend later on.

Model/Ui/OmiseConfigProvider.php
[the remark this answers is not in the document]
guzzilar: Sound like we are speaking the same thing..
(the current code is fetching Capabilities API every time when Magento displays the checkout page. Instead we (maybe) can fetch this Installment backends when the Installment payment method is enabled only)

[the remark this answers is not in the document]
jonrandy: Similar, but they should only be able to turn it on in admin IF their account has some instalment backends, or I guess we can leave the switch there, but display a message if their account has no instalment backends. If we don't do this, we're going to get spurious "instalment payments aren't working" support requests. A simple information message will avoid these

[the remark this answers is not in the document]
guzzilar: Got your point now, yeah, would be great if we can do that 
(it would also solve those legacy issues where "merchant enable 3-D Secure, IB, Alipay only from Magento side but not from Omise side" as well). _(maybe in the future PR?)_

Anyway, that is different from the "fetching Capabilities data every time Magento displays the check out page" topic.
This class will be executed only in the frontend side. And this code, `'capabilities' => $this->capabilities->get()` still need to be here to pass the instalment backends to JS file. Just need to make sure that it won't fetch the data (as it's unnecessary) if the payment method is disabled.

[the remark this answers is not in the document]
jonrandy: If we don't call it every time, and don't have the API response stored, how will we know which instalment methods to display?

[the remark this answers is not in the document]
jonrandy: Agreed

view/frontend/web/template/payment/offsite-installment-form.html
guzzilar: @jacstn Just wonder, is there any reason it won't be in this PR? Cuz this seems major one to me.
i.e. I have `installment_bay` in my Capabilities API but it doesn't show in the UI because of the hard-code.
Also some merchants might have different setting on which banks they can accept instalment?

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+                        <input type="radio" name="offsite" value="installment_first_choice" data-bind="
+                            attr: {
+                                id: getCode() + '_first_choice'
+                            },
+                            checked: omiseOffsite,
+                            click: function(data, event) { 
+                                installmentTerms(null); 
+                                return true; 
+                            };" />
+                        <label for="omise_offsite_installment_fc">
```
guzzilar: @jacstn `for` attribute has different value with the `<input>` element (`omise_offsite_installment_fc` & `getCode() + '_first_choice'`)

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,184 @@
...
+                        <input type="radio" name="offsite" value="installment_bbl" data-bind="
+                            attr: {
+                                id: getCode() + '_bbl'
+                            },
+                            checked: omiseOffsite,
+                            click: function(data, event) { 
+                                installmentTerms(null); 
+                                return true; 
+                            };" />
+                        <label for="omise_offsite_intstallment_bbl">
```
guzzilar: @jacstn typo `omise_offsite_intstallment_bbl`

guzzilar:
@jacstn 
screen shot 2562-01-07 at 07 56 22

An order's log message shows paid via Omise Internet Banking payment instead of Installment.

view/frontend/web/template/payment/offsite-installment-form.html
jonrandy: I think it needs to go in now. The number alone is essentially meaningless to the customer

view/frontend/web/template/payment/offsite-installment-form.html
jonrandy: Would be fixed by not hardcoding it, bringing the names from the `backend` data

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: @guzzilar it is knockout.js.. and I am not sure how to do it yet. 
in the same time I spoke to sales team and to product team, it is fine to hardcode it now, according to Visit we have no situation when we would limit number of installment options to our customers.

Releasing it is main priority

Controller/Callback/Offsite.php
```diff
@@ -77,7 +78,7 @@ public function execute()
             return $this->redirect(self::PATH_CART);
         }
 
-        if (! in_array($payment->getMethod(), [Alipay::CODE, Internetbanking::CODE])) {
+        if (! in_array($payment->getMethod(), [Alipay::CODE, Internetbanking::CODE, Installment::CODE])) {
```
guzzilar: @jacstn at about the line #131 of this file still saying 
`'Amount of %1 has been paid via Omise Internet Banking payment',`
(in case of successful charge).

Model/Api/Capabilities.php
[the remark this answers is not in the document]
guzzilar: cc @jacstn guess you missed this conversation.

Model/Api/Capabilities.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    private $capabilities;
+
+    public function __construct() {
+        $this->capabilities = OmiseCapabilities::retrieve();
+    }
+
+    public function read() {
```
guzzilar: @jacstn also, `read()` method doesn't reflect to its function. Would be nice to having more meaningful method name.

Model/Capabilities.php
```diff
@@ -0,0 +1,34 @@
...
+    ) {
+        $this->omise = $omise;
+        $this->capabilitiesAPI = $capabilitiesAPI;
+
+        $this->omise->defineUserAgent();
+        $this->omise->defineApiVersion();
+        $this->omise->defineApiKeys();
+    }
+
+    public function get() {
```
guzzilar: same as `Model\Api\Capabilities` comment.
The method itself doesn't describe its functionality and is ambiguous.

You could have 
```php
// Model\Api\Capabilities
class Capabilities
{
    public function retrieve()
    {
        return OmiseCapabilities::retrieve();
    }
}
```
and 
```php
// Model\Capabilities

class Capabilities
{
    public function retrieveInstallmentBackends()
    {
        $capabilities = $this->capabilitiesAPI->retrieve();
        return $capabilities->getBackends(
	    $capabilities->backendTypeIs('installment')
        );    
    }
}
```

That would be more clear when you execute it at the ConfigProvider class.
```php
public function getConfig()
{
    $storeId = $this->_storeManager->getStore()->getId();
    foreach ($this->_paymentLists->getActiveList($storeId) as $method) {
        if ($method->getCode() === OmiseInstallmentConfig::CODE) {
            return [
                'capabilities' => $this->capabilities->retrieveInstallmentBackends(),
            ];
        }
    }
}
```

etc/adminhtml/system.xml
```diff
@@ -110,6 +110,21 @@
                         <config_path>payment/omise_offline_tesco/title</config_path>
                     </field>
                 </group>
+                <group id="omise_offsite_installment" translate="label" sortOrder="160" showInDefault="1" showInWebsite="1" showInStore="1">
```
guzzilar: @jacstn `sortOrder` 160 - 180 have been used by omise_offsite_alipay already.

etc/frontend/di.xml
```diff
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
+<config
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <type name="Magento\Checkout\Model\CompositeConfigProvider">
         <arguments>
             <argument name="configProviders" xsi:type="array">
-                <item name="omise_config_provider" xsi:type="object">Omise\Payment\Model\Ui\CcConfigProvider</item>
+                <item name="cc_config_provider" xsi:type="object">Omise\Payment\Model\Ui\CcConfigProvider</item>
```
guzzilar: @jacstn wonder if it will be any duplicated-name issue.. would it be more safe to have `omise_` as prefix?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -0,0 +1,178 @@
...
+        'Magento_Checkout/js/model/url-builder',
+        'Magento_Checkout/js/model/error-processor'
+    ],
+    function (
+        $,
+        ko,
+        storage,
+        Component,
+        fullScreenLoader,
+        redirectOnSuccessAction,
```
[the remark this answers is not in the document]
guzzilar: @jacstn in case if you missed this one

view/frontend/web/template/payment/offsite-installment-form.html
guzzilar: @jacstn in case you missed this one

view/frontend/web/template/payment/offsite-installment-form.html
guzzilar: @jacstn in case you missed this one

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Retrieve assoc array of checkout configuration
+     *
+     * @return array
+     */
+    public function getConfig()
+    {
+        return [
+            'capabilities' => $this->getCapabilities(),
```
guzzilar: @jacstn do we need to return `capabilities` even the Instalment is not enabled?

view/frontend/web/template/payment/offsite-installment-form.html
guzzilar: @jacstn Then can you make sure that the list of instalment will be just only 4? Because the `installment_bay` is still missing here..

view/frontend/web/template/payment/offsite-installment-form.html
guzzilar: @jacstn in case if you missed this one

Model/Ui/CapabilitiesConfigProvider.php
jacstn: good point...

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -0,0 +1,58 @@
+<?php
+namespace Omise\Payment\Model\Ui;
+
+use Magento\Checkout\Model\ConfigProviderInterface;
+use Magento\Payment\Api\PaymentMethodListInterface;
+use Magento\Store\Model\StoreManagerInterface;
+use Omise\Payment\Model\Capabilities;
+use Omise\Payment\Model\Config\Installment as OmiseInstallmentConfig;
+
+class CapabilitiesConfigProvider implements ConfigProviderInterface
```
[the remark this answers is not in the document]
jonrandy: We can't hardcode these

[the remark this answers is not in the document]
guzzilar: @jonrandy no no, not to hard code. This just to demonstrate the data flow.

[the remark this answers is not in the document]
jacstn: Yes... had discussion few times with @jonrandy about similar stuff... and I am religious pro-push stuff to work on frontend side.. whatever it can be done on frontend we should push out from backend.. that is also where `magento 2` goes..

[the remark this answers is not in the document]
jonrandy: Yes, but we the 'stuff' and data that gets used on the front-end initially comes from the back end. There is no real point sending stuff to the front end that won't be used or still requires some initial trivial processing before use. It makes sense to send only what is needed in a form that is easily useable by the front end code

view/frontend/web/template/payment/offsite-installment-form.html
jacstn: yes.. it is missing.. and needs to be added..

Model/Ui/CapabilitiesConfigProvider.php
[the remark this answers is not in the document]
guzzilar: I also don't think frontend should process or manipulate any data if it is not necessary.
Since the data is coming from Backend, then plugin should handle the data to make it in a ready-to-use form.
That is also to limit access to an unnecessary data that it won't going to be used at the frontend as well.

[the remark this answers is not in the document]
jacstn: Ok... well if it does not threat security stuff.. it should be done on frontend.. that is just one idea in modern development.. I love it..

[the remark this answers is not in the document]
jacstn: I mean.. user's computer should calculate as much stuff as possible if it does not threat security or break architecture of few other stuff that does not apply to our situation.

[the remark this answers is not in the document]
jonrandy: Totally depends on what the processing is

Model/Capabilities.php
jacstn: not sure how future will look like, for now I changed (as we get only from `installment` backend) to getFromInstallments

So
`Capabilities::getFromInstallments` it is.

guzzilar: Uhmm.. maybe just my English.. but `getFromInstallments` give a bit different sense to me.
Like, you have `Capabilities` object. And you get `InstallmentBackends`, `InternetBankingBackends`, `CardBackends` **from** the Capabilities.

Thinking in User object case, that contains a various of user informations.
Assumes that you happen to want to get a specific user's firstname to display on the page.
You surely won't call `$user->getFromFirstname` but `$user->getFirstname`.

Same thing to Capabilities object.

guzzilar: Or maybe in a more related-case. 
Assume that you wanted to display a purchased-list of a specific user at the user-order-history page. You want to know how many times that this specific user purchased stuff from the store.
You probably go with
```php
foreach($user->purchasedList() as $item) {
    echo "$item->amount - $item->purchased_date";
}
```
or
```php
foreach($user->getPurchasedList() as $item) {
    echo "$item->amount - $item->purchased_date";
}
```

But `getFromPurchasedList()` wouldn't make sense for sure
```php
foreach($user->getFromPurchasedList() as $item) {
    echo "$item->amount - $item->purchased_date";
}
```

jonrandy: I'm totally with @guzzilar on this. Reusing the name `capabilities` in a different context from the way it is reference both in `omise-php` and the Capabilities API itself simply introduces unnecessary confusion. Better to stick with the terminology originating from the API.

We are interested in retrieving the properties of a payment `backend` from within the `capabilities` of the merchant account. In this case, this backend is of the `instalment` type. @guzzilar 's suggestions for naming seem very appropriate and communicate well what they do in the context of the Capabilities API

As for `Capabilities::getFromInstallments` or `getFromInstallmentBackend`, to me the names feel linguistically awkward and they depart from conventions in other code that were drawn from those adopted in the API response itself

jonrandy: I noticed the same confusing use of `capabilities` in the `getConfig` method - instalment backends are being called 'capabilities'

Model/Capabilities.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->omise->defineUserAgent();
+        $this->omise->defineApiVersion();
+        $this->omise->defineApiKeys();
+    }
+
+    /**
+     * @return array
+     */
+    public function retrieveInstallmentBackends() {
+        return $this->capabilitiesAPI->getBackends();
```
guzzilar: @jacstn how about 
```php
public function retrieveInstallmentBackends()
{
    return $this->capabilitiesAPI->getInstallmentBackends();
}
```

Because now `$this->capabilitiesAPI->getBackends()` returns `InstallmentBackends`. 
If one day if you need to retrieve `InternetbankingBackends` there won't be a place for that.

view/frontend/web/css/styles.css
```diff
@@ -0,0 +1,96 @@
...
+#omise_offsite_internetbankingForm img.fc, #omise_offsite_installmentForm img.fc {
+    background: url('../images/fc.svg') #007bc4;
+}
+/** KBank **/
+#omise_offsite_internetbankingForm .kbank, #omise_offsite_installmentForm .kbank {
+    background: rgb(247, 226, 226);
+}
+#omise_offsite_internetbankingForm img.kbank, #omise_offsite_installmentForm img.kbank {
+    background: url('../images/kbank.svg')rgb(247, 226, 226);
+}
```
guzzilar: @jacstn one breakline here please

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -0,0 +1,202 @@
...
+                        click: placeOrder,
+                        attr: {title: $t('Place Order')},
+                        css: {disabled: !isPlaceOrderActionAllowed()},
+                        enable: (getCode() == isChecked())">
+                    <span data-bind="i18n: 'Place Order'"></span>
+                </button>
+            </div>
+        </div>
+    </div>
+</div>
```
guzzilar: @jacstn one brekline here pls

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @return array
+     */
+    public function getConfig()
+    {
+        $listOfActivePaymentMethods = $this->_paymentLists->getActiveList($this->_storeManager->getStore()->getId());
+        foreach ($listOfActivePaymentMethods as $method) {
+            if ($method->getCode() === OmiseInstallmentConfig::CODE) {
+                return [ 
+                    'capabilities' => $this->capabilities->retrieveInstallmentBackends()
```
guzzilar: @jacstn I would like to suggest to rename this one to something else as well (`installment_backends`, `installment_capabilities` would make it).
The reason is that `window.checkoutConfig` variable in Javascript file is sharing across the checkout page.
Meaning that if in the future you want to retrieve a list of banks that accept Internet Banking then it will come back to the naming issue.

view/frontend/web/css/styles.css
guzzilar: can we have one breakline here?

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -0,0 +1,47 @@
+<?php
+namespace Omise\Payment\Model\Ui;
+
+use Magento\Checkout\Model\ConfigProviderInterface;
+use Magento\Payment\Api\PaymentMethodListInterface;
+use Magento\Store\Model\StoreManagerInterface;
+use Omise\Payment\Model\Capabilities;
+use Omise\Payment\Model\Config\Installment as OmiseInstallmentConfig;
+
+class CapabilitiesConfigProvider implements ConfigProviderInterface
```
guzzilar: So since this class is providing configs specifically for Installment, should we rename this one to just `InstallmentConfigProvider`?

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return array
+     */
+    public function getConfig()
+    {
+        $listOfActivePaymentMethods = $this->_paymentLists->getActiveList($this->_storeManager->getStore()->getId());
+        foreach ($listOfActivePaymentMethods as $method) {
+            if ($method->getCode() === OmiseInstallmentConfig::CODE) {
+                return [ 
+                    'installment_backends' => $this->capabilities->retrieveInstallmentBackends()
```
guzzilar: just want to note it out that we are sending out 3 unnecessary properties. From Capabilities API response,
```json
{
  "installment_bay": {
    "type": "installment",
    "currencies": [
      "thb"
    ],
    "allowed_installment_terms": [
      3,
      4,
      6,
      9,
      10
    ],
    "amount": {
      "min": 2000,
      "max": 100000000
    }
  }
}
```

In this pull request, it uses only `allowed_installment_terms` and `_id` (special attribute generated by Omise-PHP).
We don't use `type`, `currencies`, and `amount` at the moment.

Would be nice if we limit an access to only what we will use in the frontend.

jacstn: will use them in the future... when we will have correct values in amount.
than I will also use currencies, now, it is also hardcoded.

Model/Ui/CapabilitiesConfigProvider.php
jacstn: not sure... 

Capabilities, we provide here capabilities for all payment methods in the future?

PR #149, opened 2018-09-06
Offsite refactoring, changed Offsite names to APM

jacekstan:
Tesco lotus uses the same functionality as all other methods except CC, and tesco is not offsite method, so we need to change names for more fitting to offsite and offline payments.

So now now since Offsite functionality is common also for Offline payments like Tesco Lotus, and future payments like Installment this PR is proposal is to have 2 descriptions.
OmiseAPMXXXXXX - for `offsite` and `offline` methods
and OmiseCCXXXXX - for `CC` credit card related methods

#### 1. Objective

This PR is necessary as for offline _tesco lotus_ functionality to use the same parts of code as offsite payments (alipay internet banking).

#### 2. Description of change

Just names of classes,

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.

**✏️ Details:**

To properly test it you need to make payments with offsite Alipay, Internet Banking, and with Credit Card method.

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Gateway/Request/PaymentAPMBuilder.php
+++ b/Gateway/Request/PaymentAPMBuilder.php
@@ -4,11 +4,11 @@ namespace Omise\Payment\Gateway\Request;
 use Magento\Framework\UrlInterface;
 use Magento\Payment\Gateway\Helper\SubjectReader;
 use Magento\Payment\Gateway\Request\BuilderInterface;
-use Omise\Payment\Model\Config\Offsite\Alipay;
-use Omise\Payment\Model\Config\Offsite\Internetbanking;
-use Omise\Payment\Observer\OffsiteInternetbankingDataAssignObserver;
+use Omise\Payment\Model\Config\Alipay;
+use Omise\Payment\Model\Config\Internetbanking;
+use Omise\Payment\Observer\InternetbankingDataAssignObserver;
 
-class PaymentOffsiteBuilder implements BuilderInterface
+class PaymentAPMBuilder implements BuilderInterface
 {
 
     /**
@@ -60,7 +60,7 @@ class PaymentOffsiteBuilder implements BuilderInterface
                 break;
             case Internetbanking::CODE:
                 $paymentInfo[self::SOURCE] = [
-                    self::SOURCE_TYPE => $method->getAdditionalInformation(OffsiteInternetbankingDataAssignObserver::OFFSITE)
+                    self::SOURCE_TYPE => $method->getAdditionalInformation(InternetbankingDataAssignObserver::OFFSITE)
                 ];
                 break;
         }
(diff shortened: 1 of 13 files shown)


Gateway/Request/PaymentAPMBuilder.php
```diff
@@ -4,11 +4,11 @@
...
 use Magento\Payment\Gateway\Request\BuilderInterface;
-use Omise\Payment\Model\Config\Offsite\Alipay;
-use Omise\Payment\Model\Config\Offsite\Internetbanking;
-use Omise\Payment\Observer\OffsiteInternetbankingDataAssignObserver;
+use Omise\Payment\Model\Config\Alipay;
+use Omise\Payment\Model\Config\Internetbanking;
+use Omise\Payment\Observer\InternetbankingDataAssignObserver;
 
-class PaymentOffsiteBuilder implements BuilderInterface
+class PaymentAPMBuilder implements BuilderInterface
```
guzzilar: @jacekstan (optional) just quick look, APM already stands for Alternative Payment Method (maybe, not 100% sure lol)
PaymentAPM is a bit redundant. How about just `APMBuilder`?

Model/Config/Alipay.php
```diff
@@ -1,5 +1,5 @@
 <?php
-namespace Omise\Payment\Model\Config\Offsite;
+namespace Omise\Payment\Model\Config;
```
guzzilar: @jacekstan I got the point that we want to make a generic class that share its functionality between `offsite` and `offline` payment methods as now we have `Gateway/Request/PaymentAPMBuilder.php`.
But not sure why we need to move the classes out from `Offsite` folder here?

jacekstan: That is very simple structure with small classes that have 1 file only, I don't see the reason why it cannot be in one folder at first place.

Now I will add tesco, and will have to create `offline` folder, I just don't see the point to complicate that structure for those few configs.

Gateway/Request/PaymentAPMBuilder.php
jacekstan: other builders have Payment word in the same folder..
almost all of them...
Actually now thinking.. should we add Payment to one missing name, or remove from all of them...
and that not an error.. you have a lot in different sectors of life.. that you repeat element from abbrev. just to show what that entire name construction is about. Even if any letter of that abbreviation stands for already used word in that name (not sure if you know what I mean.)

guzzilar: Shall we remove all then?
I was prefixing it as `Payment` with no strong reason, just to clarify that the class is for payment.
But after more methods coming, I kinda have a second thought that probably we can just remove it all, since everything in `Gateway` folder will always relate to the Payment function.

PR #150, opened 2018-09-07
Mitigate issue caused by multiple hits on `return_uri`

jonrandy:
#### 1. Objective

An issue arose whereby payments were being recorded multiple times in Magento for an order, even though only one payment had been made through Omise. This appeared to be a strange edge case where the `return_uri` was somehow getting hit multiple times. We need to add some kind of check to make sure we only update the payment on the invoice if we have not previously done so.

#### 2. Description of change

A small check was added to each of the controllers hit by `return_uri` to make sure that the invoice was not already updated if we are attempting to update it.

The code to check the invoice state was added to the Omise Order model (`isInvoicePaid`)

#### 3. Quality assurance

Payments were retested for each type. The case of repeated hits to `return_uri` was also simulated to check the result was correct.

#### 4. Impact of the change

No observable changes to functionality or UI, but hopefully no more recurrences of this issue

#### 5. Priority of change

Normal

#### 6. Additional Notes

🛵

### DIFF
--- a/src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitealipayController.php
+++ b/src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitealipayController.php
@@ -37,9 +37,11 @@ class Omise_Gateway_Callback_ValidateoffsitealipayController extends Omise_Gatew
 
         if ($charge->isSuccessful()) {
             $invoice = $order->getInvoice($payment->getLastTransId());
+            $transId = $payment->getLastTransId();
 
-            $order->markAsPaid(
-                $payment->getLastTransId(),
+            // Make sure to avoid marking invoice paid more than once
+            if (!$order->isInvoicePaid($transId)) $order->markAsPaid(
+                $transId,
                 Mage_Sales_Model_Order::STATE_PROCESSING,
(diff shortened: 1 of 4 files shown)


src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitealipayController.php
```diff
@@ -37,9 +37,11 @@ public function indexAction()
 
         if ($charge->isSuccessful()) {
             $invoice = $order->getInvoice($payment->getLastTransId());
+            $transId = $payment->getLastTransId();
 
-            $order->markAsPaid(
-                $payment->getLastTransId(),
+            // Make sure to avoid marking invoice paid more than once
+            if (!$order->isInvoicePaid($transId)) $order->markAsPaid(
```
guzzilar: @jonrandy how about we just add this condition inside the `markAsPaid` method instead?

jonrandy:
I did originally, but thought better of it

jonrandy:
My thinking was if partial payment came along, it would need probably to
come back out

src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitealipayController.php
jonrandy: see below

PR #151, opened 2018-09-10
Switch to consistent naming of `private` and `protected` class members

jonrandy:
#### 1. Objective

I noticed that we were randomly enforcing a naming convention of using an underscore prefix for `private` and `protected` class members. This should be enforced right across the plugin

#### 2. Description of change

Renamed all affected class members so that they now have the required underscore prefix

#### 3. Quality assurance

Re-tested all plugin functionality after making the changes

#### 4. Impact of the change

Now easier to see at a glance whether a class member is public or not

#### 5. Priority of change

Normal

#### 6. Additional Notes

👾👾👾

### DIFF
--- a/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
+++ b/src/app/code/community/Omise/Gateway/Block/Form/Cc.php
@@ -11,7 +11,7 @@ class Omise_Gateway_Block_Form_Cc extends Mage_Payment_Block_Form
      */
     protected function _prepareLayout()
     {
-        if ($this->isApplicable()) {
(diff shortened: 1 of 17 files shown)


guzzilar:
@jonrandy @jacekstan  prefixed `_` is not recommended in PSR-2 which the one that Magento applied with. 
https://www.php-fig.org/psr/psr-2/#42-properties

As we are trying to follow the platform code standard, I think we should just leave it without `_` prefix.

guzzilar:
(also, we might get into some issue when we submit out extension to their store, as Magento is kinda strict on coding-standard)

PR #123, opened 2018-06-11
MAG2x: Add a capability to remember user's credit card and use it for a second-time purchase

guzzilar:
### 1. Objective

One of a powerful credit card feature is to "remember user's credit card".
This would make the payment process with credit card even faster by allowing users to save their credit card information with their Magento user account so users don't need to re-fill the credit card form again when they come to a second purchase.

> Note: According to a security-best-practice, in fact, we do not save any informations of a user's card into a merchant's store. Instead, we do exchange the user's credit card information with Omise service via Omise Customer API at the very first moment when user purchase a product with `save card` checked.
> Then, Omise will send back a Customer Object with a `customer.id`, which Omise-Magento plugin will only keep this `customer.id` to use to refer back to a specific Customer Object and its card.
 
### 2. Description of change

- New option at the checkout page, credit card payment method, `remember this card`.
- Users will be able to save their credit card information and retrieve it back for a second-time purchase.
- User will be able to charge with an existing card (1-click payment)

### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: v7.1.16.
- **Platform version**: Magento Open Source v2.2.5.

**✏️ Details:**
**3.1). Make sure that user can create a charge with a guest-cart (not logged-in user) properly (on both auth only and auth and capture payment).**
- As a normal process, user must be able to checkout using Omise Credit Card payment either authorise only or authorise and capture payment action.

_[1] The credit card payment form should be shown as normal._
screen shot 2561-08-17 at 17 46 59

_[2] And user must be able to check out an order without having any problem with a guest-cart._
screen shot 2561-08-17 at 17 47 42

_[3] Transaction will be appeared at the admin Magento order detail page._
screen shot 2561-08-17 at 17 48 47

_[4] Omise received the charge request and processed it correctly (authorise and capture)._
screen shot 2561-08-17 at 17 52 46

_[5] Omise received the charge request and processed it correctly (authorise only)._
screen shot 2561-08-17 at 17 55 34

**3.2). Make sure that user can create a charge using their Magento account (on both auth only and auth and capture payment).**
- Create a new account on the Magento store
- Logged in to the store and proceed the check out process as normal.
- As a normal process, user must be able to checkout using Omise Credit Card payment either authorise only or authorise and capture payment action.

_[1] Create a new account on the test-store._
screen shot 2561-08-17 at 18 14 44

_[2] Logged in to the store and check out an order as normal._
screen shot 2561-08-17 at 18 17 38

[...]

PR #153, opened 2018-09-18
MAG2x: Changing long array syntax to the short syntax one.

guzzilar:
> ⚠️ **Note, this pull request requires PR #123 to be merged first.**

#### 1. Objective

screen shot 2561-09-18 at 14 54 14

There are code standard provided by Magento that all extensions have to follow in order to submit the code to their Magento Marketplace.

This pull request is aiming to update one of the standard (array syntax) to a proper one that Magento requires.

#### 2. Description of change

- Update all long-syntax-array to the short one.

#### 3. Quality assurance

1. Install https://github.com/magento/marketplace-eqp.
2. Execute the command  `vendor/bin/phpcs /path/to/your/extension --standard=MEQP2`.
3. Search for a warning message `Short array syntax must be used to define arrays`.

**Before change, we should see these ones**
screen shot 2561-09-18 at 14 54 14

**After change, we should see none of them**
screen shot 2561-09-18 at 15 01 40

#### 4. Impact of the change

Nothing in a user perspective, but it will make our extension be able to pass the Magento Code Quality Standard and able to submit to the Magento Marketplace.

#### 5. Priority of change

Normal

#### 6. Additional Notes

-

### DIFF
--- a/Test/Unit/Model/Api/ObjectTest.php
+++ b/Test/Unit/Model/Api/ObjectTest.php
@@ -98,7 +98,7 @@ class MockObject extends Object
 class MockOmiseObject implements \ArrayAccess
 {
     // Store the attributes of the object.
-    protected $_values = array();
+    protected $_values = [];
 
     public function refresh()
     {
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -77,7 +77,7 @@ class Offsite extends Action
             return $this->redirect(self::PATH_CART);
         }
 
-        if (! in_array($payment->getMethod(), array(Alipay::CODE, Internetbanking::CODE))) {
+        if (! in_array($payment->getMethod(), [Alipay::CODE, Internetbanking::CODE])) {
             $this->invalid($order, __('Invalid payment method. Please contact our support if you have any questions.'));
 
             return $this->redirect(self::PATH_CART);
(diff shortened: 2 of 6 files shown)


Test/Unit/Model/Api/ObjectTest.php
```diff
@@ -98,7 +98,7 @@ public function reload()
 class MockOmiseObject implements \ArrayAccess
 {
     // Store the attributes of the object.
-    protected $_values = array();
+    protected $_values = [];
```
jonrandy: Protected var name starting with `_`

guzzilar: @jonrandy Thanks
Code-refactoring PR will be coming in soon!

jacekstan: @guzzilar no necessary to change in pr only few places.. change all in entire plugin..

guzzilar: @jacekstan yup, so changing entire plugin will be coming in an another pull request once `tesco-lotus` feature is merged (or maybe before) (I updated array syntax & opened this pull request just to continue from a conversation at the previous PR #123)

jacekstan: so why this PR *now* is necessary?
and why there is problem to follow all plugin with those changes?

Issue #89, opened 2017-08-24
System not sent email order to customer when make a payment via omise 


A-Kits:
I found a issue about email transaction.
When customer make a payment via omise, A email order was not sent to customer.
This issue happen only Omise payment other method such as Paypal, Direct Deposit not have tis issue.
plugin Omise ver 1.12 Magento ver.1.9.3.4 

Thank

guzzilar:
@A-Kits Hi

Can you check at your order that has email transaction issue if that order status was set to 'pending' or 'processing'?

A-Kits:
Hi
Order status was sent as processing and has a notification " email order
was not sent" on head of order page.

guzzilar:
@A-Kits Ok thanks. I'll check on it.

A-Kits:
Hi, Do you  have any updates?
Did issue happen with other user ?

guzzilar:
@A-Kits Hi,
This ticket is still in progress.
However, so far, I haven't heard this issue from other users.

A-Kits:
Hi 
Do you test this plugin with your magento system? 
I concern that this issue happen because omise plugin or because of my magento system.
Anyway for my website  this issue happened with only omise payment.

A-Kits:
Do you have any update for this issue?

guzzilar:
@A-Kits Hi, I kinda found one spot that may relate to this issue. Working on it now (but I guess Im gonna need a bit more time).

bokoko:
im facing the same problem here. My magento version is 2.1.0. The other payment option can send e-mail confirmation order just fine, but the email do not send when user pick omise payment option.

thank you very much

guzzilar:
Hi @bokoko thanks for your report.
I'm adding tag `status: Need investigation` here as we are trying to reproduce the issue but still can't figure it out why this problem occurred to you (we've tested and email has been sent properly).

But there has to be something behind.. will let you know once we have any updates❗️ 

By the way, I'm guessing you are using Omise-Magento v2.2 right?

Jamjann:
Hi everyone.Now I'm using Magento 1.9. and also face this problems too.Do you have any updates?

Jamjann:
Hi everyone.Now I'm using Magento 1.9. and also face this problems.Do you have any updates?

keeratita:
I think this issue should be closed due to it was fixed on Omise-Magento version 1.16.

PR #159, opened 2018-10-25
Adds required code to fix order emails not being sent in all 'redirect' scenarios

jonrandy:
#### 1. Objective

Fixes the issue with order emails not being sent out. It appears that any payment using a method involving a redirect (3D secure, Alipay, Internet Banking) was not sending out an order email after payment.

**Related information**:
Related issue(s): #89 

#### 2. Description of change

Added `$order->sendNewOrderEmail();` in all required places

#### 3. Quality assurance

Tested on Magento 1.9.3.9

Tested all payment methods available in Omise, both in `authorise` and `authorise and capture` modes. Also tested with webhooks on/off to make sure all cases handled correctly (no duplicate emails etc.)

Emails should look like this:
screenshot 2018-11-07 at 12 35 25

This [plugin](https://github.com/aschroder/Magento-SMTP-Pro-Email-Extension) is particularly useful for setting up email in Magento so you can test easily. It also logs all emails

#### 4. Impact of the change

Order emails should now be sent out correctly in all cases

#### 5. Priority of change

Normal

#### 6. Additional Notes

This issue also affects Magento 2 - hopefully the fix will be similar.

### DIFF
--- a/src/app/code/community/Omise/Gateway/Model/Event/Charge/Complete.php
+++ b/src/app/code/community/Omise/Gateway/Model/Event/Charge/Complete.php
@@ -54,6 +54,7 @@ class Omise_Gateway_Model_Event_Charge_Complete
 
         if ($isPendingOrReview && ($charge->isSuccessful() || $charge->isAwaitCapture())) {
             $order->getPayment()->accept();
+            $order->sendNewOrderEmail();
             return $order->save();
         }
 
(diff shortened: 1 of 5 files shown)


guzzilar:
@jonrandy Can we have a screenshot of an email template that will be sent out from `$order->sendNewOrderEmail();`?

Issue #167, opened 2018-12-03
setup:di:compile displays error when install on Magento 2.3


nisapus:
Hi,

I've tried to setup Omise with Magento 2.3, but found the error when run setup:di:compile.

$ php bin/magento setup:di:compile
Compilation was started.
Repositories code generation... 1/7 [====>-----------------------]  14% < 1 sec 70.5 MiBPHP Fatal error:  Cannot use 'Object' as class name as it is reserved in /var/www/html/magento-23/vendor/omise/omise-magento/Model/Api/Object.php on line 7

Fatal error: Cannot use 'Object' as class name as it is reserved in /var/www/html/magento-23/vendor/omise/omise-magento/Model/Api/Object.php on line 7

FYI, test on Ubuntu 18.04 with PHP 7.2

jacstn:
Thanks @nisapus .... I am on it.. it will be released with next release.
I will let you know soon how to quickly fix it.

jacstn:
closing as fixed

PR #168, opened 2018-12-11
Make plugin compatible with PHP 7.2

jacstn:
#### 1. Objective

Make plugin compatible with PHP 7.2

**Related information**:
#167 

#### 2. Description of change

Changed class name from `Object` to `BaseObject` - can't use general `Object` name in php 7.2

#### 3. Quality assurance

**🔧 Environments:**
- **Platform version**: Magento CE 2.3.
- **PHP version**: 7.2.

**✏️ Details:**
Use PHP 7.2 to install fresh plugin in Magento 2.3
Note that Magento below `2.3` is not compatible with `PHP 7.2`

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -100,7 +100,7 @@ class Offsite extends Action
         try {
             $charge = $this->charge->find($charge_id);
 
-            if (! $charge instanceof \Omise\Payment\Model\Api\Object) {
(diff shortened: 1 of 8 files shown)


jacstn:
@guzzilar no flush cache is necessary, composer will wipe that file (I am almost 100% sure).
When upgrading manually than it is like copying files in any other case.. old files will remain.

Issue #173, opened 2019-01-15
Upgrading from version 2.4 to version 2.5. Class not found 'OmiseCapabilities'


BrainTurner:
Dear Omise,

I have upgraded both omise-php(2.91) and omise magento(2.5) and get the following error
PHP message: PHP Fatal error:  Uncaught Error: Class 'OmiseCapabilities' not found in /var/www/clients/client1/web1/web/vendor/omise/omise-magento/Model/Api/Capabilities.php:13
Stack trace:
#0 /var/www/clients/client1/web1/web/vendor/magento/framework/ObjectManager/Factory/Compiled.php(60): Omise\Payment\Model\Api\Capabilities->__construct()
#1 /var/www/clients/client1/web1/web/vendor/magento/framework/ObjectManager/Factory/Compiled.php(150): Magento\Framework\ObjectManager\Factory\Compiled->create('Omise\\Payment\\M...')
#2 /var/www/clients/client1/web1/web/vendor/magento/framework/ObjectManager/Factory/Compiled.php(79): Magento\Framework\ObjectManager\Factory\Compiled->get('Omise\\Payment\\M...')
#3 /var/www/clients/client1/web1/web/vendor/magento/framework/ObjectManager/Factory/Compiled.php(150): Magento\Framework\ObjectManager\Factory\Compiled->create('Omise\\Payment\\M...')
#4  

I have flush cache, setup upgrade, di:compile, restarted varnish, pushed files to content server and still the same message.

My system setup is:
OS: centos7
PHP: php70w-fpm
Cache: varnish and redis
Magento 2.2.5
CDN: AWS Cloudfront

I have been waiting for this upgrade after you moudle upgraded to 2.4 and had so many problems and now i cannot use it and i cannot downgrade back to 2.4 ?

Please advise

jacstn:
hello.

Sorry about that.. 

Please also update your `omise-php` to the newest version.

https://github.com/omise/omise-php/releases/tag/v2.11.0

guzzilar:
@BrainTurner Hi, Omise-Magento v2.5 now requires for Omise-PHP v2.11.0.
Could you share us how did you upgrade Omise-PHP and Omise-Magento? If it is by composer, you could run: `composer require omise/omise-php:2.11.0` at your Magento root folder to upgrade Omise-PHP to the latest version.

Let me know if there is still a problem after upgrading Omise-PHP to v2.11.0.

BrainTurner:
[...]

PR #178, opened 2019-01-16
Fix: Order can't be completed when choosing COD

jacstn:
#### 1. Objective

There is error when choosing Cash on delivery payment method.

Reason is that our extension does not get required additional data when choosing different than `omise` payment methods. 

Issue has been described in different report: 

This is fix for that issue. [#173](https://github.com/omise/omise-magento/issues/173)

#### 2. Description of change

We need to check if there is `additional_information` passed in payment data array in displaying payment success page.

#### 3. Quality assurance

**🔧 Environments:**

- **Omise plugin version**: Omise-Magento 2.5.
- **PHP version**: 7.0.16.

**✏️ Details:**

Make payment with Cash On Delivery as payment option.

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Observer/TescoPaymentObserver.php
+++ b/Observer/TescoPaymentObserver.php
@@ -39,19 +39,15 @@ class TescoPaymentObserver implements ObserverInterface
     {
         $order   = $observer->getEvent()->getOrder();
         $payment = $order->getPayment();
-        
-        if (!$payment) {
+
+        if (!$payment->getAdditionalData('payment_type') !== 'bill_payment_tesco_lotus') {
             return $this;
         }
 
-        $paymentData = $payment->getData();
-
-        if ($paymentData['additional_information']['payment_type'] !== 'bill_payment_tesco_lotus') {
-            return $this;
-        }
+        $paymentData   = $payment->getData();
 
         $amount        = number_format($paymentData['amount_ordered'], 2) . ' ' . $order->getOrderCurrency()->getCurrencyCode();
-        $barcodeHtml   = $this->_helper->convertTescoSVGCodeToHTML($paymentData['additional_information']['barcode']);
+        $barcodeHtml   = $this->_helper->convertTescoSVGCodeToHTML($payment->getAdditionalData('barcode'));
         $storeName     = $this->_scopeConfig->getValue('trans_email/ident_sales/name', \Magento\Store\Model\ScopeInterface::SCOPE_STORE);
         $storeEmail    = $this->_scopeConfig->getValue('trans_email/ident_sales/email', \Magento\Store\Model\ScopeInterface::SCOPE_STORE);
         $customerEmail = $order->getCustomerEmail();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+### [v2.6 _(Jan 15, 2019)_](https://github.com/omise/omise-magento/releases/tag/v2.6)
+
+### 👾 Bug Fixes
+- Model\Ui\CapabilitiesConfigProvider: returns an empty array instaed of null when Installment payment method is disabled. (PR [#174](https://github.com/omise/omise-magento/pull/174))
+
 ### [v2.5 _(Jan 14, 2019)_](https://github.com/omise/omise-magento/releases/tag/v2.5)
 
 ### ✨ Highlights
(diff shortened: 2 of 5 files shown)


Observer/TescoPaymentObserver.php
```diff
@@ -46,7 +46,7 @@ public function execute(Observer $observer)
 
         $paymentData = $payment->getData();
 
-        if ($paymentData['additional_information']['payment_type'] !== 'bill_payment_tesco_lotus') {
+        if (!isset($paymentData['additional_information']) || !isset($paymentData['additional_information']['payment_type']) || $paymentData['additional_information']['payment_type'] !== 'bill_payment_tesco_lotus') {
```
guzzilar: @jacstn I just looked at `module-sales/Model/Order/Payment.php` and ``module-sales/Model/Order/Payment/Info.php`. 
Seems we can use 👇 instead.
```php
if (! $payment->hasAdditionalInformation('payment_type') || $payment->getAdditionalInformation('payment_type') !== 'bill_payment_tesco_lotus ') {
    // ...
}
```

Or could even be
```php
if ($payment->getAdditionalInformation('payment_type') !== 'bill_payment_tesco_lotus ') {
    // ...
}
```
`getAdditionalInformation($key)` will return `null` if the key doesn't exist.

jacstn: yes.. but that `$payment->getAdditionalInformation('barcode')`
returned always null for some reason.

guzzilar: Oh, it might be because it contains URL?
Maybe escape HTML once before set to additionalInformation would help?

But well, that's another topic.
For now, I think that `!isset($paymentData['additional_information'])` is unncessary but if you want to have it here, then you might want to consider to have it at `Block/Checkout/Onepage/Success/TescoAdditionalInformation.php` as well? 

Cuz in that class it also checks similar thing but just only
```php
if (!isset($paymentData['additional_information']['payment_type']) || $paymentData['additional_information']['payment_type'] !== 'bill_payment_tesco_lotus') {
    // ...
}
```

jacstn: the barcode is SVG raw code... but your suggestion might be a good lead.

also, I wanted to be consistant.. in the file.. it is only one way to approach, it might look better, to reduce code, but it is only one way to do it.

since I will have to get that `additional_data` in a raw array to get a barcode, then why not to do it all in the same way (treat additional data).

I will try to get that `barcode` using that method, will try to escape string.

jacstn:
well, this is why we have PR description, right? 
the title might be short.

I will work on the title, but referring to any code `variables` in PR title might be too much? :)

Observer/TescoPaymentObserver.php
guzzilar: I mean I think that `!isset($paymentData['additional_information'])` is unnecessary to check cuz after looked at Magento classes, `additional_information` is a default key generated by Magento, so it's should always presented there.

But if you want to check it, then maybe we should also update `Block/Checkout/Onepage/Success/TescoAdditionalInformation.php` class, cuz right now that class does not check for `!isset($paymentData['additional_information'])` case.

guzzilar:
@jacstn We can't reach to the right pull request's description if its title isn't clear anyway.
Just I've found that choosing COD method is not a real root-cause, root-cause is choosing any payment method that does not set `payment_type`.

And `Fix for script error` is a bit ambiguous (if I have the same problem with other payment method, I would start searching for _Undefined index: payment_type_ in GitHub's search first - _but it might just only me, to be fair_)

But anyway, your call.
I'm just throwing an idea here :)

guzzilar:
Another idea throwing out: 
"Fix 'Undefined index: payment_type' error when making an order with any payment method while Omise-Magento is enabled"
😂

Observer/TescoPaymentObserver.php
jacstn: ok.. yes.. that might be too much

jacstn:
I have followed your suggestion regarding commit message (there is only 1 commit)
But regarding PR description i prefer high level.

Observer/TescoPaymentObserver.php
```diff
@@ -39,19 +39,15 @@ public function execute(Observer $observer)
     {
         $order   = $observer->getEvent()->getOrder();
         $payment = $order->getPayment();
-        
-        if (!$payment) {
```
jacstn: this is from _Payment observer_ I don't expect $payment be ever null.

jacstn: well actually it is trigged after `sales_order_place_after` event.. still don't expect it can be `null`

PR #179, opened 2019-01-17
Fix: Order can't be completed when choosing other payment methods than Omise

jacstn:
#### 1. Objective

There is error when choosing Cash on delivery payment method.

Reason is that our extension does not get required additional data when choosing different than `omise` payment methods. 

Issue has been described in different report: 

This is fix for that issue. [#173](https://github.com/omise/omise-magento/issues/173)

#### 2. Description of change

We need to check if there is `additional_information` passed in payment data array in displaying payment success page.

#### 3. Quality assurance

**🔧 Environments:**

- **Omise plugin version**: Omise-Magento 2.5.
- **PHP version**: 7.0.16.

**✏️ Details:**

Make payment with Cash On Delivery as payment option.

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Observer/TescoPaymentObserver.php
+++ b/Observer/TescoPaymentObserver.php
@@ -39,19 +39,15 @@ class TescoPaymentObserver implements ObserverInterface
     {
         $order   = $observer->getEvent()->getOrder();
         $payment = $order->getPayment();
-        
-        if (!$payment) {
+
+        if ($payment->getAdditionalData('payment_type') !== 'bill_payment_tesco_lotus') {
             return $this;
         }
 
-        $paymentData = $payment->getData();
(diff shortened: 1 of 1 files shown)


Observer/TescoPaymentObserver.php
```diff
@@ -39,19 +39,15 @@ public function execute(Observer $observer)
     {
         $order   = $observer->getEvent()->getOrder();
         $payment = $order->getPayment();
-        
-        if (!$payment) {
```
jacstn: I don't expect that in this callback after order is compleated $payment object will be ever null

PR #180, opened 2019-01-17
Fix: Display correctly barcode from tesco lotus on checkout success page.

jacstn:
#### 1. Objective

Display barcode fix. 
There is error that barcode is not displayed on tesco lotus page.
This PR deliver fix for that functionality.

#### 2. Description of change

Just renamed function to get proper array value.

array value we are pointing in:
`TescoAdditionalInformation.php:39` is named `tesco_code_image`,
but we try to get by calling `getTescoCodeUrl`:
`tesco_code_url`

We removed `img` tag, because we display Tesco barcode using SVG code.
In that PR also was changed array key represents Tesco image displayed in checkout success page.
`tesco_code_image` => `tesco_barcode_svg`.

and getter function was renamed to `getTescoBarcodeSvg`.

#### 3. Quality assurance

**🔧 Environments:**

- **Omise plugin version**: Omise-Magento 2.6.
- **PHP version**: 7.0.29.

**✏️ Details:**

Make Tesco Payment,
See if barcode is displayed properly.

#### 4. Impact of the change

Solve issue

#### 5. Priority of change

High,

#### 6. Additional Notes

None

### DIFF
--- a/view/frontend/templates/checkout/onepage/success/tesco_additional_info.phtml
+++ b/view/frontend/templates/checkout/onepage/success/tesco_additional_info.phtml
@@ -1,5 +1,5 @@
 
 <p><?= __('To finish payment, print the code below and bring it to make payment in any Tesco Lotus Store. Code is valid for the next <b>24 hours only</b>. After that time your order will be canceled.'); ?></p>
-<p style="margin:.5rem 0"><?= $block->escapeHtml(__('Amount to pay'))?>: <strong><?= $block->getOrderAmount();?></b></strong>
-<div><?= $block->getTescoCodeImage() ?></div>
+<p style="margin:.5rem 0"><?= $block->escapeHtml(__('Amount to pay')) ?>: <strong><?= $block->getOrderAmount() ?></b></strong>
+<div><?= $block->getTescoBarcodeSvg() ?></div>
 <p><button class="action secondary" onclick="window.print ? window.print() : alert('Print not available')"><?= __('Print')?></button></p>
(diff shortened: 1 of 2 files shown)


view/frontend/templates/checkout/onepage/success/tesco_additional_info.phtml
```diff
@@ -1,5 +1,5 @@
 
 <p><?= __('To finish payment, print the code below and bring it to make payment in any Tesco Lotus Store. Code is valid for the next <b>24 hours only</b>. After that time your order will be canceled.'); ?></p>
 <p style="margin:.5rem 0"><?= $block->escapeHtml(__('Amount to pay'))?>: <strong><?= $block->getOrderAmount();?></b></strong>
-<p><img src="<?= $block->escapeUrl($block->getTescoCodeUrl()) ?>" /></p>
+<div><?= $block->getTescoCodeImage() ?></div>
```
guzzilar: @jacstn just wanna point out that actually looking at this change without understanding the context seems a bit strange. 
While the code is changed to `getTescoCodeImage()` but we removed `</img>` tag.

jacstn: after digging deeper into code, we still display in HTML - SVG code.

SVG is converted to HTML only when it is necessary - before sending it in email.

`img` with `src` was necessary when we were using `URL` to `SVG` file.
I renamed it to `getTescoBarcodeSvg` and corresponding array key to: `tesco_barcode_svg`

PR #182, opened 2019-01-24
Manual Capture Functionality

jacstn:
#### 1. Objective

Allow Merchant to use Manual Capture functionality from Magento 2 Admin Panel

#### 2. Description of change

When a user pays by Credit Card using `authorize_only` method than Merchant has to go to `Omise.co` Dashboard to manually capture payment.

This Pull Request allow to manual capture payment during creating an invoice directly from Magento 2 Admin Panel.

There has been added information about existing `charge_id` when there is created request body of Credit Card Payment.
If `charge_id` already exists then all request is to Manual Capture payment.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

To manually test that Pull Request:

-  Make sure you have set up in Admin Panel -> Store -> Configuration -> Sales -> Payment Methods -> Omise -> Credit Card Solution -> Payment Action as *Authorize Only*
- Make order with CC Payment.
- Go to Omise Dashboard and make sure that last payment you did is _authorized only_
- Go to Admin Panel, in Orders select last order and create an invoice using _Submit Invoice_ button, make sure you have selected _Capture Online_ option.
image

- After creating invoice go to Omise Dashboard and check if Payment is successfully _captured_.

#### 4. Impact of the change

N/A

#### 5. Additional Notes

N/A

### DIFF
--- a/Gateway/Http/Client/AbstractPayment.php
+++ b/Gateway/Http/Client/AbstractPayment.php
@@ -7,7 +7,7 @@ use Magento\Payment\Gateway\Http\TransferInterface;
 use Omise\Payment\Model\Api\Charge as ApiCharge;
 use Omise\Payment\Model\Omise;
 
-class Payment implements ClientInterface
+abstract class AbstractPayment implements ClientInterface
 {
     /**
      * Client request status represented to successful request step.
@@ -49,8 +49,15 @@ class Payment implements ClientInterface
         ApiCharge $apiCharge,
         Omise     $omise
     ) {
-        $this->omise     = $omise;
+        $this->omise = $omise;
         $this->apiCharge = $apiCharge;
+    }
+
+    protected function setupOmiseLib()
+    {
+        $this->omise->defineUserAgent();
+        $this->omise->defineApiVersion();
+        $this->omise->defineApiKeys();
     }
 
     /**
@@ -58,21 +65,5 @@ class Payment implements ClientInterface
      *
      * @return \Omise\Payment\Model\Api\Charge|\Omise\Payment\Model\Api\Error
      */
-    public function placeRequest(TransferInterface $transferObject)
-    {
-
-        $this->omise->defineUserAgent();
-        $this->omise->defineApiVersion();
-        $this->omise->defineApiKeys();
-        
-        $transferObjectBody = $transferObject->getBody();
-
-        // if charge_id already exists than action is 'manual capture'
-        if (isset($transferObjectBody[self::CHARGE_ID])) {
-            $charge = $this->apiCharge->find($transferObjectBody[self::CHARGE_ID]);
-            return [self::CHARGE => $charge->capture()];
-        }
-
-        return [self::CHARGE => $this->apiCharge->create($transferObjectBody)];
-    }
+    abstract public function placeRequest(TransferInterface $transferObject);
 }
--- a/Gateway/Http/Client/AbstractPayment.php
+++ b/Gateway/Http/Client/AbstractPayment.php
@@ -45,19 +45,19 @@ abstract class AbstractPayment implements ClientInterface
      */
     protected $apiCharge;
 
+    /**
+     * @param \Omise\Payment\Model\Api\Charge $apiCharge
+     * @param \Omise\Payment\Model\Omise $omise;
+     */
     public function __construct(
         ApiCharge $apiCharge,
         Omise     $omise
     ) {
-        $this->omise = $omise;
-        $this->apiCharge = $apiCharge;
-    }
+        $omise->defineUserAgent();
+        $omise->defineApiVersion();
+        $omise->defineApiKeys();
 
-    protected function setupOmiseLib()
-    {
-        $this->omise->defineUserAgent();
-        $this->omise->defineApiVersion();
-        $this->omise->defineApiKeys();
+        $this->apiCharge = $apiCharge;
     }
 
     /**
--- a/Gateway/Http/Client/CCPayment.php
+++ b/Gateway/Http/Client/CCPayment.php
@@ -11,8 +11,6 @@ class CCPayment extends AbstractPayment
      */
     public function placeRequest(\Magento\Payment\Gateway\Http\TransferInterface $transferObject)
     {
-        $this->setupOmiseLib();
-
         $transferObjectBody = $transferObject->getBody();
 
         // if charge_id already exists than action is 'manual capture'
--- a/Gateway/Http/Client/CCPayment.php
+++ b/Gateway/Http/Client/CCPayment.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Omise\Payment\Gateway\Http\Client;
+
+class CCPayment extends AbstractPayment
+{
+    /**
+     * @param  \Magento\Payment\Gateway\Http\TransferInterface $transferObject
+     *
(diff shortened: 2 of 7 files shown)


Gateway/Http/Client/Payment.php
```diff
@@ -50,10 +48,19 @@ public function __construct(
...
         $this->omise->defineUserAgent();
         $this->omise->defineApiVersion();
         $this->omise->defineApiKeys();
+        
+        $transferObjectBody = $transferObject->getBody();
+
+        // if charge_id already exists than action is 'manual capture'
+        if ($chargeID = $transferObjectBody['charge_id']) {
+            $charge = $this->apiCharge->find($chargeID);
+            return ['charge' => $charge->capture()];
```
guzzilar: This doesn't feel a right place and I feel the code smell..
`Omise\Payment\Gateway\Http\Client\Payment` is a class that shared between all payment methods (Alipay, Credit Card, Installment, Internet Banking, Tesco Bill Payment).
I think there might be a better place than having having a condition for a specific case like card-capture at this class (cannot tell at the moment, I will have a look a bit more).

Gateway/Request/CreditCardBuilder.php
```diff
@@ -21,15 +22,24 @@ class CreditCardBuilder implements BuilderInterface
     public function build(array $buildSubject)
     {
         $payment = SubjectReader::readPayment($buildSubject);
-        $method  = $payment->getPayment();
+        $method = $payment->getPayment();
+
+        $paymentData = null;
```
guzzilar: I think this one should be empty array than `null`

Gateway/Request/CreditCardBuilder.php
```diff
@@ -21,15 +22,24 @@ class CreditCardBuilder implements BuilderInterface
...
+            ];
+        } else {
+            $paymentData = [
+                self::CARD => $method->getAdditionalInformation(CreditCardDataObserver::TOKEN)
             ];
         }
 
-        return [ self::CARD => $method->getAdditionalInformation(CreditCardDataObserver::TOKEN) ];
+        //add information about charge_id, if charge ID already exists than it is 'manual capture' request.
+        $paymentData[self::CHARGE_ID] = $method->getAdditionalInformation(CreditCardDataObserver::CHARGE_ID);
```
guzzilar: Just continue from https://github.com/omise/omise-magento/pull/182/files#r250871230
This case is for `capture` only right? Do we need this to be assigned to `paymentData` variable (that contains `self::CARD` and `self::CUSTOMER`)?

guzzilar: 👆 Not sure if I can explain it properly but now the output from this method in the charge-capture case would be either
```php
$paymentData = [
    'self::CARD' = "",
    'self::CHARGE_ID' = "chrg_test_51qwe**",
]
```
or
```php
$paymentData = [
    'self::CUSTOMER' = "",
    'self::CARD' = "",
    'self::CHARGE_ID' = "chrg_test_51qwe**",
]
```
which we don't seem to need either `CARD` or `CUSTOMER` to be presented there in the array for the charge-capture case.

jacstn: this part was the same as it was... 
I just instead of doing instant return I did the first assign to a variable, then do something with that variable, and then return it. Before it was instant return.

guzzilar: Sorry, I mean, do you need `self::CUSTOMER`, `self::CARD` to be presented in the array if it's just a case for charge-capture.
It was instant return for 2 cases, charge with a token, and charge with a saved-card from a specific Omise customer. But for the case of charge-capture, maybe we just need 
```php
return [ self::CHARGE_ID = $method->getAdditionalInformation(CreditCardDataObserver::CHARGE_ID) ];
```

jacstn: yes.. to manual capture we need only `charge_id`

jacstn: but I need to send in `paymentData` info about `charge_id` even if it is `null` all the time, because later there is decision made what that request is about based on `charge_id` status.

Gateway/Request/CreditCardBuilder.php
jacstn: it might be `null` IMO, because later there is only `setting` data into it. no other operations, this var is always `re-written`.

As a case in the past when there is later only `=` operation on var than it was set to `null` in the as declaration.

Gateway/Http/Client/Payment.php
jonrandy: +1

Gateway/Http/Client/Payment.php
```diff
@@ -50,10 +48,19 @@ public function __construct(
...
+        $transferObjectBody = $transferObject->getBody();
+
+        // if charge_id already exists than action is 'manual capture'
+        if (isset($transferObjectBody['charge_id'])) {
+            $charge = $this->apiCharge->find($transferObjectBody['charge_id']);
+            return ['charge' => $charge->capture()];
+        }
 
-        return ['charge' => $this->apiCharge->create($transferObject->getBody())];
+        return ['charge' => $this->apiCharge->create($transferObjectBody)];
```
jonrandy: In this function, `'charge'` has been hardcoded whereas in the file below a constant has been used. Should we be consistent?

Gateway/Http/Client/Payment.php
jacstn: it is payment, payment might be: create `charge` or  payment `capture`.

we have following payment actions:
`authorize`, `authorize_capture` or `captured` previously authorized

Until last time this Class was responsible only for `authorize` or `authorize_capture`, because `capture` was not implemented yet. Now it is.
Yes it is shared by all methods, and it is fine to put it here, in the future I believe there will be more methods that will do `authorize` and than `capture` payment action.

Gateway/Http/Client/CCPayment.php
```diff
@@ -0,0 +1,26 @@
...
+class CCPayment extends AbstractPayment
+{
+    /**
+     * @param  \Magento\Payment\Gateway\Http\TransferInterface $transferObject
+     *
+     * @return \Omise\Payment\Model\Api\Charge|\Omise\Payment\Model\Api\Error
+     */
+    public function placeRequest(\Magento\Payment\Gateway\Http\TransferInterface $transferObject)
+    {
+        $this->setupOmiseLib();
```
guzzilar: @jacstn I guess we can move this line to its parent's constructor method?

Gateway/Http/Client/AbstractPayment.php
```diff
@@ -25,6 +23,18 @@ class Payment implements ClientInterface
      */
     const PROCESS_STATUS_FAILED = 'failed';
 
+    /**
+     *
+     * @var string
+     */
+    const CHARGE_ID = 'charge_id';
```
guzzilar: @jacstn I see this `CHARGE_ID` constant has been declared several times in multiple places (here, the request builder class, the observer class). Should we reduce the amount of this duplication?

Maybe we don't need this constant here and instead, just user from the builder class?

jacstn: this will be the subject of another PR, I will think about what single place should be owning that const.

PR #169, opened 2019-01-08
Instalment payments for Magento 1

jonrandy:
#### 1. Objective

Add the 'Instalment Payment' payment option, featuring options defined by the Capability API response for the merchant's account

#### 2. Description of change

- Switches to using newer API version
- Fixes all other payment methods affected by the above change
- Adds relevant code to make Capabilities features for Omise-PHP available in Magento
- Adds new Instalment payment method utilising the Capabilities feature above

#### 3. Quality assurance

Checked all existing payment methods still working.
Checked new instalment method working (albeit with the horrible issue with 'cancelled' bank pages)

Tested on
- **Platform version**: Magento CE 1.9.3
- **Omise plugin version**: Latest (kind of custom right now, but will copy in latest version before merging this PR)
- **PHP version**: 5.6.33

#### 4. Impact of the change

image

The store will now have instalment payment capability

Admin interface:
screenshot 2019-02-15 at 12 47 56

Success:
screenshot 2019-02-15 at 12 53 09

Failure:
screenshot 2019-02-15 at 12 56 54

#### 5. Priority of change

Normal

PR #184, opened 2019-02-15
Deal with unknown instalment providers (merge after #169)

jonrandy:
#### 1. Objective

Since it is possible that a new instalment payment provider that the code is not aware of could appear in the Capabilities API response, we need to handle this situation gracefully (show a generic icon and some sensible text). Hopefully this situation never occurs, but we should be prepared for it.

#### 2. Description of change

Modifies HTML template and CSS to facilitate the changes mentioned above. Also adds a generic SVG 'provider' icon

#### 3. Quality assurance

Kinda tricky to test. Manually hacking the HTML of the page to reflect an 'unknown' provider is the quickest way - unless you want to actually insert a new provider into the Capability API response somehow.

#### 4. Impact of the change

Instalments section of payments at checkout will now look okay should an unknown instalment provider creep into the capability API response

screenshot 2019-02-15 at 16 23 09

#### 5. Priority of change

Normal, High or Immediate.

#### 6. Additional Notes

B, G, C#


guzzilar:
@jonrandy can you provide screenshots of this change at the PR description?
Thanks.

src/skin/frontend/base/default/omise/gateway/css/style.css
```diff
@@ -45,53 +47,53 @@
     background: #4e2e7f;
 }
 .banks-list .logo-scb {
-    background: url('../images/scb.svg') #4e2e7f;
+    background: url('../images/scb.svg') #4e2e7f !important;
```
guzzilar: @jonrandy generally I think we should avoid using `!important` when it is not necessary. 
Some approaches like
```css
.banks-list .omise-logo-wrapper > div.logo-scb {
    background: url('../images/scb.svg') #4e2e7f;
}
```
Would give the same result but reserve `!important` for another case, such 3rd-party developer / designer to customize it.

Btw, i think we may not need to have these changes at all. 
Let me explain in the next comment.

guzzilar:
@jonrandy While this approach works as expected. I have another idea to propose.
How about we just don't need to display those `unknown` banks at all?

The reason is about UX. For example, if Omise supports 2 new banks that we don't have it in the plugin before, it would give a result with 7 items:
```
• [ FiC logo ]     First Choice
• [ BAY logo ]     Bay
• [ SCB logo ]     SCB
• [ KTB logo ]     KTB
• [ BBL logo ]     BBL
• [ Unknown logo ] Unknown installment provider
• [ Unknown logo ] Unknown installment provider
```
In the UI/UX perspective, those 2 unknown items would give zero benefit to user while it also give an urgent sense of `broken` application and needed to be fixed (on both user and merchant perspectives). 

Instead, if we filter those out and display only the known-banks that the plugin has information with, even though Omise provides new banks in the future, it would not affect to the plugin and give us some times to patch the plugin accordingly.

src/skin/frontend/base/default/omise/gateway/css/style.css
guzzilar: explained here https://github.com/omise/omise-magento/pull/184#issuecomment-464032191

jonrandy:
I actually thought of this after I already made this fix. Thought I'd just see what you guys think. Even better would be if the English names actually came from the Capability API - then we could add new ones (albeit with no pretty logo) without the merchants having to upgrade the plugin. Seeing as we have no names coming from the API though, not displaying providers we do not have names for is probably the most user-friendly, support-minimising way to go

jonrandy:
@jacstn What are your thoughts on this? Display or not display the 'unknown' instalment provider?

jacstn:
@jonrandy @guzzilar well.. I would keep unknown banks giving more options to a user, to be honest, but I would change that item to:

`Installment bank provider (installment_first_choice)`
as an example from `JSON` would do.

It is not a crime to show some code information even if it is variable.

Also, we will be reacting very fast (I hope) in case of new installment providers. So that will be very temporary I hope.

jonrandy:
Yup - adding the provider 'code' in some way is another option

guzzilar:
> Also, we will be reacting very fast (I hope) in case of new installment providers. So that will be very temporary I hope.

Note that even we can patch and release within 1-2 days, but imagine that we have 100 merchants using this version, there will be 100 queries emails to the support team asking about this issue. 
Also note that, especially for big enterprise - corporate. To deploy / upgrade one patch, there might be a lot of procedures to proceed depending on those company's structure and could takes up to many days. Or if it hits on Friday, then it would at least take 3-6 days to patch (1-2 days from us, 2 days for weekend, 1-2 day for them to patch (in best scenario))

Not saying that we should completely not to display those list, but might need to consider on an impact to those parties as well. (if we can patch that fast, then why not let user wait 1 or 2 more days for a complete and not-a-bug-like version from us - is it really matter to display a new provider at the checkout page right away after Omise provided a new one - and if it is worthy to display an incomplete list just for 1-2 days before the patch)

jonrandy:
Would a new instalment provider be added and activated on a merchant account without them being informed first?

jacstn:
yes... I still think we should display unknown companies, but yes... maybe in a smarter way...

PR #186, opened 2019-03-28
Delete Saved Credit Card Information

jacstn:
#### 1. Objective

This Pull Request deliver functionality to delete saved credit card.

#### 2. Description of change

This PR adds to the customer profile new tab called "Omise Payment Methods".

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.29.

**✏️ Details:**

To manually test this feature:

*When no card is saved*
- Login to your magento 2 customer profile
- Open customer profile and check if on tab _Omise Payment Methods_ you have information that no payment card is available.
Screenshot 2019-03-28 at 18 24 09

*When you have saved credit card*
- Make sure you are logged in as customer
- Check if in _Omise Payment Methods_ you have listed all saved credit cards.
- Delete credit card.
- Make sure deleted card is no more listed on the list.

#### 4. Impact of the change
New Tab was added to Magento 2 Customer Profile.

Screenshot 2019-03-28 at 17 44 25


Controller/Methods/Cards.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Dispatch request
+     *
+     * @return \Magento\Framework\Controller\ResultInterface|ResponseInterface
+     * @throws \Magento\Framework\Exception\NotFoundException
+     */
+    public function execute()
+    {
+        $resultPage = $this->pageFactory->create();
+        $resultPage->getConfig()->getTitle()->set(__('Stored Payment Methods'));
```
jonrandy: The table on the page only contains cards, and has a column heading reinforcing this fact. Shouldn't the page title be "Saved Payment Cards", "Saved Credit/Debit Cards", "Stored Payment Cards", or "Stored Credit/Debit Cards"?

Controller/Methods/Delete.php
```diff
@@ -0,0 +1,128 @@
...
+        return $this->_redirect('omise/methods/cards');
+    }
+
+    /**
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createSuccessMessage()
+    {
+        $this->messageManager->addSuccessMessage(
+            __('Stored Payment Method was successfully removed')
```
jonrandy: Should this refer to card?

view/frontend/layout/customer_account.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
+<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
+    <body>
+        <referenceBlock name="customer_account_navigation">
+            <block class="Magento\Customer\Block\Account\SortLinkInterface" name="customer-account-navigation-my-credit-cards-link">
+                <arguments>
+                    <argument name="path" xsi:type="string">omise/methods/cards</argument>
+                    <argument name="label" xsi:type="string" translate="true">Omise Payment Methods</argument>
```
jonrandy: I think we shouldn't put 'Omise' here - I think it should be the same as whatever the title of the page is. The customer does not care or maybe even know what Omise is, and the store is highly unlikely to have more than one card payment handler active that could cause confusion

jacstn: it should be omise, it is not general place, other payment methods will create their own subpage here.

Controller/Methods/Cards.php
jacstn: there will be other payment methods here in the future.

view/frontend/layout/customer_account.xml
jonrandy: I strongly disagree. Have you ever seen a store with more than one card payment provider active at once?

Controller/Methods/Delete.php
jacstn: it does not matter.. does it? it is just confirmation, fully describing what just happened.

view/frontend/layout/customer_account.xml
jonrandy: I have never once seen a store that stores card details mention the name of the provider they use for payments

Controller/Methods/Cards.php
jacstn: BTW it is also exactly the same way of displaying as magento does it in `module_vault` `cards_list.phtml`

jonrandy: So other payment methods will be in different tables on the same page? The table specifically only refers to cards. This contradiction should be resolved one way or the other

Controller/Methods/Delete.php
jonrandy: Well maybe, but if there are only cards, the wording should reflect that. Also, the capitalisation needs fixing

Controller/Methods/Cards.php
jonrandy: @guzzilar Thoughts?

Controller/Methods/Delete.php
jonrandy: A well designed UX should always remove all unnecessary questions from a user's mind

jonrandy: @guzzilar Thoughts?

view/frontend/layout/customer_account.xml
jonrandy: @guzzilar Thoughts?

Controller/Methods/Cards.php
guzzilar: @jacstn what would those other payment methods could possibly be (in the future)?

Controller/Methods/Delete.php
guzzilar: Just wonder why "delete a saved card" should return `Stored Payment Method was successfully removed`, not "The card was successfully removed".

view/frontend/layout/customer_account.xml
guzzilar: Agreed with @jonrandy. `Omise Payment Method` doesn't give any hints to users that they can look up their saved card here. (imagine that you are a customer, decided to save a card with www.starbucks.co then, you want to delete your card later on, where or which menu will you click to see your cards?)

Controller/Methods/Delete.php
jacstn: credit / depit card is payment method.

Controller/Methods/Cards.php
jacstn: don't know, but I have strong believes in omise.. any ideas about code?

Controller/Methods/Delete.php
guzzilar: But you don't delete payment method, you delete card, right?

Controller/Methods/Cards.php
guzzilar: So I'm not sure what does `Stored Payment Methods` exactly mean?
About code, I think we should avoid over-design if possible unless we clearly see what is coming after and know what it is for. Or else it gives a confusion as for user experience.

Controller/Methods/Delete.php
```diff
@@ -0,0 +1,128 @@
+<?php
+namespace Omise\Payment\Controller\Methods;
+
+use Magento\Framework\App\Request\Http;
+use Magento\Framework\Exception\NotFoundException;
+
+class Delete extends \Magento\Framework\App\Action\Action
```
guzzilar: @jacstn `Controller/Methods/Delete` is a bit ambiguous, don't you think?

Controller/Methods/Delete.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createSuccessMessage()
+    {
+        $this->messageManager->addSuccessMessage(
+            __('Stored Payment Method was successfully removed')
+        );
+
+        return $this->_redirect('omise/methods/cards');
```
guzzilar: @jacstn I believe `createSuccessMessage` method shouldn't be taking care of redirection.
Maybe move out to a method that executing it?

Controller/Methods/Delete.php
```diff
@@ -0,0 +1,128 @@
...
+        $cardId = $this->getCardID($request);
+
+        if ($cardId === null) {
+            return $this->createErrorResponse(self::WRONG_TOKEN);
+        }
+
+        try {
+            $this->customer->deleteCard($cardId);
+        } catch (\Exception $e) {
+            return $this->createErrorResponse(self::ACTION_EXCEPTION);
```
guzzilar: @jacstn how about also send `$e->getMessage()` for the reason of failure?

view/frontend/layout/customer_account.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
+<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
+    <body>
+        <referenceBlock name="customer_account_navigation">
+            <block class="Magento\Customer\Block\Account\SortLinkInterface" name="customer-account-navigation-my-credit-cards-link">
+                <arguments>
+                    <argument name="path" xsi:type="string">omise/methods/cards</argument>
```
guzzilar: why do we need `methods` in the URL here.. 🤔

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,57 @@
...
+                    <th scope="col" class="col actions">&nbsp;</th>
+                </tr>
+            </thead>
+            <tbody>            
+<?php foreach($cards['data'] as $card): ?>
+                <tr>
+                    <td><?php echo ' **** **** **** ' . $card['last_digits']; ?></td>
+                    <td><?php echo $card['expiration_month']; ?>/<?php echo $card['expiration_year']; ?></td>
+                    <td><?php echo $card['brand'] ?></td>
+                    <td><a id="cc-delete" href="<?php echo $block->getDeleteLink($card)?>"><?= /* @noEscape */ __('Delete') ?></a></td>
```
guzzilar: @jacstn id is referencing to only a specific one unique element. In this case, should be `class`.

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,57 @@
...
+                </tr>
+<?php endforeach; ?>
+            </tbody>
+        </table>
+    </div>
+<?php elseif (count($cards['data']) === 0): ?>
+    <div class="message info empty"><span><?= /* @noEscape */ __('You have no stored payment methods.') ?></span></div>
```
guzzilar: @jacstn the code is trying to list a card out, if card data = 0, then shouldn't it says "no card" instead?

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,57 @@
...
+                    <td><?php echo ' **** **** **** ' . $card['last_digits']; ?></td>
+                    <td><?php echo $card['expiration_month']; ?>/<?php echo $card['expiration_year']; ?></td>
+                    <td><?php echo $card['brand'] ?></td>
+                    <td><a id="cc-delete" href="<?php echo $block->getDeleteLink($card)?>"><?= /* @noEscape */ __('Delete') ?></a></td>
+                </tr>
+<?php endforeach; ?>
+            </tbody>
+        </table>
+    </div>
+<?php elseif (count($cards['data']) === 0): ?>
```
guzzilar: @jacstn maybe
```html
<?php if (!empty($cards['data'])): ?>
    <!-- list cards -->
<?php else: ?>
    <!-- display none -->
<?php endif; ?>
```
is enough to check?

Controller/Methods/Cards.php
jonrandy: +1

Avoid putting unnecessary questions in the head of the user

Controller/Methods/Delete.php
jonrandy: 'Card' is definitely much clearer here

view/frontend/templates/cards.phtml
jonrandy: I commented on this too. Obviously didn't save. Damned crappy Starbucks internet

Controller/Methods/Delete.php
jacstn: after delete we do refresh page.
We could think in the future about `ajax` solution

Controller/Methods/Delete.php
jacstn: that is how it is handled by other magento modules.

view/frontend/layout/customer_account.xml
jacstn: as omise payment methods... I might change?

Controller/Manage/Delete.php
```diff
@@ -0,0 +1,128 @@
...
+        return $this->_redirect('omise/manage/cards');
+    }
+
+    /**
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createSuccessMessage()
+    {
+        $this->messageManager->addSuccessMessage(
+            __('Saved Credit/Debit Card data was successfully removed')
```
jonrandy: Saved credit/debit card data was successfully removed

jonrandy: Capitalisation

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,56 @@
+<script type="text/javascript">// <![CDATA[
+require([
+        'jquery',
+        'Magento_Ui/js/modal/confirm'
+    ],
+    function($, confirmation) {
+         $('.omise-cc-delete').on('click', function(event){
+             event.preventDefault();
+                 confirmation({
+             content: 'Are you sure you want to delete card?',
```
jonrandy: Does this get translated?

jonrandy: Are you sure you want to delete this card?

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,56 @@
...
+                    <td><?php echo $card['expiration_month']; ?>/<?php echo $card['expiration_year']; ?></td>
+                    <td><?php echo $card['brand'] ?></td>
+                    <td><a class="omise-cc-delete" href="<?php echo $block->getDeleteLink($card) ?>"><?=/* @noEscape */__('Delete')?></a></td>
+                </tr>
+<?php endforeach;?>
+            </tbody>
+        </table>
+    </div>
+<?php else: ?>
+    <div class="message info empty"><span><?=/* @noEscape */__('You have no stored card.')?></span></div>
```
jonrandy: You have no stored cards.

guzzilar:
@jacstn note, this PR's commits looks off. Maybe some accidents when it got rebased or merged somehow?

Model/Customer.php
```diff
@@ -128,6 +128,20 @@ public function addCard($cardToken)
...
+     * @param  string $cardToken
+     *
+     *
+     * @see    https://github.com/omise/omise-php/blob/master/lib/omise/OmiseCardList.php
+     */
+    public function deleteCard($cardToken)
+    {
+        $card = $this->customerAPI->cards()->retrieve($cardToken);
+        $card->destroy();
+    }
```
guzzilar: @jacstn wonder if we should really retrieve a card here before delete.
Why not let the controller retrieve a card first then leave the delete method as just to delete?

Controller/Manage/Delete.php
```diff
@@ -0,0 +1,128 @@
...
+            $this->errorsMap[$errorCode]
+        );
+
+        return $this->_redirect('omise/manage/cards');
+    }
+
+    /**
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createSuccessMessage()
```
guzzilar: @jacstn Wonder why we wouldn't let `Block\Customer\Cards` handle all display-related things (as it seems to be designed specifically for this job). Or is there any limitation on Magento structure?

view/frontend/layout/omise_manage_cards.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+
+<page
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../../../../../lib/internal/Magento/Framework/View/Layout/etc/page_configuration.xsd">
```
guzzilar: @jacstn is it safe to use relative path here? `../../../../../../../lib/internal/Magento/Framework/View/Layout/etc/page_configuration.xsd`

Would it be any issue when it comes to install the plugin via manual (placing folder at `app/code/Omise/Payment` and composer (inside `vendor` folder).

Controller/Manage/Delete.php
jacstn: @guzzilar not sure, as I used nomenclature from that module before, and there were reported issues about the misunderstanding that by customers. All strings before you and @jonrandy asked me to change due misunderstanding were from that module.

jacstn: No... I used the magento approach from `Magento_Vault`

Model/Customer.php
jacstn: I don't understand what you want..

view/frontend/layout/omise_manage_cards.xml
jacstn: I have copied that from module in `vendor`, so probably it would work, but playing it safe, I did change it to absolute.

Controller/Cards/DeleteAction.php
```diff
@@ -0,0 +1,128 @@
...
+            self::WRONG_REQUEST    => __('Wrong request.'),
+            self::ACTION_EXCEPTION => __('Deletion failure. Please try again.'),
+        ];
+    }
+
+    /**
+     * Dispatch request
+     *
+     * @return Magento\Framework\Controller\ResultInterface|Magento\Framework\App\ResponseInterface
+     * @throws NotFoundException
```
guzzilar: Just wonder, in what case this would throw NotFoundException?

Controller/Cards/ListAction.php
```diff
@@ -0,0 +1,64 @@
...
+use Magento\Customer\Model\Session;
+use Magento\Framework\App\Action\Context;
+use Magento\Framework\App\RequestInterface;
+use Magento\Framework\App\ResponseInterface;
+use Magento\Framework\Exception\NotFoundException;
+
+class ListAction extends \Magento\Framework\App\Action\Action
+{
+    /**
+     * @var Session
```
guzzilar: `\Magento\Customer\Model\Session` maybe? (just see other classes documented here as a full path namespace)

Controller/Cards/DeleteAction.php
jacstn: i copied it from existing magento code, You are right, will remove that comment.

Controller/Cards/ListAction.php
```diff
@@ -0,0 +1,64 @@
...
+    public function dispatch(RequestInterface $request)
+    {
+        if (!$this->customerSession->authenticate()) {
+            $this->_actionFlag->set('', 'no-dispatch', true);
+        }
+        return parent::dispatch($request);
+    }
+
+    /**
+     * Dispatch request
```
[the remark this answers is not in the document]
jacstn: yes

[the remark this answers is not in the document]
jacstn: removed.

Controller/Cards/ListAction.php
```diff
@@ -0,0 +1,46 @@
...
+     * {@inheritdoc}
+     */
+    public function execute()
+    {
+        $resultPage = $this->pageFactory->create();
+        $resultPage->getConfig()->getTitle()->set(__('Stored Credit/Debit Cards'));
+
+        return $resultPage;
+    }
+}
```
guzzilar: @jacstn I see now you completely removed the dispatch method, so we don't need to check if `!$this->customerSession->authenticate()` anymore or it has been done by Magento as default already?

Block/Customer/Cards.php
```diff
@@ -0,0 +1,46 @@
...
+        parent::__construct($context, $data);
+        $this->customer = $customer;
+    }
+
+    /**
+     * @return  array|null
+     */
+    public function getCards()
+    {
+        return !$this->customer->getMagentoCustomerId() || !$this->customer->getId() ? null : $this->customer->cards(['order' => 'reverse_chronological']);
```
jonrandy: `reverse_chronological` feels like it should be a constant

jacstn: If you search for that in magento code, they use that as string, we do too. This is better as `constant` but better would be to do it as seperated PR and find proper place to put that constant, and replace that in other places.

Controller/Cards/DeleteAction.php
```diff
@@ -0,0 +1,120 @@
...
+     * @param int $errorCode
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createErrorResponse($errorCode)
+    {
+        $this->messageManager->addErrorMessage(
+            $this->errorsMap[$errorCode]
+        );
+
+        return $this->_redirect('omise/cards/listaction');
```
jonrandy: Path feels like it should be a constant - same goes for the other action paths

Controller/Cards/DeleteAction.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * @return Magento\Framework\App\ResponseInterface
+     */
+    private function createSuccessMessage()
+    {
+        $this->messageManager->addSuccessMessage(
+            __('Saved credit/debit card data was successfully removed')
```
jonrandy: Remove the word 'data'

Controller/Cards/DeleteAction.php
jacstn: could be.. but most paths are not used as constants in Magento 2. check for instance `Magento_Vault\Controller\DeleteAction::createSuccessMessage`

view/frontend/templates/cards.phtml
```diff
@@ -0,0 +1,57 @@
+<script type="text/javascript">// <![CDATA[
+require([
+        'jquery',
+        'Magento_Ui/js/modal/confirm'
+    ],
+    function($, confirmation) {
+         $('.omise-cc-delete').on('click', function(event){
+             event.preventDefault();
+                 confirmation({
```
jonrandy: Indentation seems wrong/weird

jacstn: looks like, I used right click/Format Document in VC. @jonrandy what happens when you right click and format?

Block/Customer/Cards.php
jonrandy: OK, if they're doing it that way

Controller/Cards/DeleteAction.php
jonrandy: Yeah, but you've repeated it twice right in the same place. Maintainability would be better?

view/frontend/templates/cards.phtml
jonrandy: By hand generally. Rarely if ever use autoformatters

PR #191, opened 2019-05-08
Refactoring - removed unnecessary redirections

jacstn:
#### 1. Objective
This PR removes unnecessary redirections that are handled already in JavaScript file generated here:
`view/frontend/templates/cards.phtml:14`

#### 2. Description of change

Just removed a few lines of code.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.1.19.

**✏️ Details:**

To manually test this PR please try to delete card in Customer panel.
Observe if the card disappears from the list and success message is displayed.

### DIFF
--- a/Controller/Cards/DeleteAction.php
+++ b/Controller/Cards/DeleteAction.php
@@ -54,19 +54,19 @@ class DeleteAction extends \Magento\Framework\App\Action\Action
     {
         $request = $this->_request;
         if (!$request instanceof Http) {
-            return $this->createErrorResponse(self::WRONG_REQUEST);
+            return $this->createErrorMessage(self::WRONG_REQUEST);
         }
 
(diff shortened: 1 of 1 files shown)


guzzilar:
Actually also maybe method name? 
Those `createErrorResponse` and `createSuccessMessage` are exactly doing the same thing but the method name giving the different impression. Maybe it should be `createErrorMessage`?

Controller/Cards/DeleteAction.php
```diff
@@ -76,15 +74,12 @@ public function execute()
 
     /**
      * @param int $errorCode
-     * @return Magento\Framework\App\ResponseInterface
```
guzzilar: Still has this DocBlock at `createSuccessMessage`. No need to delete?

PR #196, opened 2019-05-15
Display Installment terms

jacstn:
#### 1. Objective

This PR added functionality to display installment terms.

#### 2. Description of change

Hardcoded interest rates and monthly minimum payments have been added on a per-provider basis. The installment options presented to the customer at the checkout page have been modified to include monthly payment amounts and information/disclaimers about the interest rates. Payment options resulting in monthly payments below the defined minimums are automatically removed from those available for selection

image

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

To test this PR make payment with installment, observe if there are amounts properly displayed.
Check if minimum amounts are respected.

ie. If you choose order value for 3k THB than 12 months option should not be displayed, due to the minimum installment amount is not met in any provider.
More info about minimum installment amounts [here](https://www.omise.co/installment-payment)

Check both options for absorbing rates by Merchant and Customer.

Check if TH translated texts regarding this functionality are displayed.

#### 4. Impact of the change

 N/A

#### 5. Priority of change

High

#### 6. Additional Notes

N/A

### DIFF
--- a/Model/Api/Capabilities.php
+++ b/Model/Api/Capabilities.php
@@ -30,7 +30,7 @@ class Capabilities extends BaseObject
      *
      * @return bool
      */
-    public function isZeroInterests() {
+    public function isZeroInterest() {
         return $this->capabilities['zero_interest_installments'];
     }
 }
--- a/Model/Api/Capabilities.php
+++ b/Model/Api/Capabilities.php
@@ -15,7 +15,7 @@ class Capabilities extends BaseObject
 
     /**
      * Get Installment capabilities array from Omise-PHP
-     * 
+     *
      * @return array
      */
     public function getInstallmentBackends()
@@ -23,5 +23,14 @@ class Capabilities extends BaseObject
         return $this->capabilities->getBackends(
             $this->capabilities->makeBackendFilterType('installment')
         );    
+    }
+
+    /**
+     * Get information about zero interest installments
+     *
+     * @return bool
+     */
+    public function isZeroInterests() {
+        return $this->capabilities['zero_interest_installments'];
     }
 }
--- a/Model/Capabilities.php
+++ b/Model/Capabilities.php
@@ -40,8 +40,8 @@ class Capabilities
     /**
      * @return bool
      */
-    public function isZeroInterests()
+    public function isZeroInterest()
     {
-        return $this->capabilitiesAPI->isZeroInterests();
+        return $this->capabilitiesAPI->isZeroInterest();
     }
 }
--- a/Model/Capabilities.php
+++ b/Model/Capabilities.php
@@ -36,4 +36,12 @@ class Capabilities
     {
         return $this->capabilitiesAPI->getInstallmentBackends();
     }
+
+    /**
+     * @return bool
+     */
+    public function isZeroInterests()
+    {
+        return $this->capabilitiesAPI->isZeroInterests();
+    }
 }
--- a/Model/Ui/CapabilitiesConfigProvider.php
+++ b/Model/Ui/CapabilitiesConfigProvider.php
@@ -38,7 +38,8 @@ class CapabilitiesConfigProvider implements ConfigProviderInterface
         foreach ($listOfActivePaymentMethods as $method) {
             if ($method->getCode() === OmiseInstallmentConfig::CODE) {
                 return [ 
-                    'installment_backends' => $this->capabilities->retrieveInstallmentBackends()
+                    'installment_backends' => $this->capabilities->retrieveInstallmentBackends(),
+                    'is_zero_interest' => $this->capabilities->isZeroInterests()
                 ];
             }
         }
--- a/Model/Ui/CapabilitiesConfigProvider.php
+++ b/Model/Ui/CapabilitiesConfigProvider.php
@@ -39,7 +39,7 @@ class CapabilitiesConfigProvider implements ConfigProviderInterface
             if ($method->getCode() === OmiseInstallmentConfig::CODE) {
                 return [ 
                     'installment_backends' => $this->capabilities->retrieveInstallmentBackends(),
-                    'is_zero_interest' => $this->capabilities->isZeroInterests()
+                    'is_zero_interest' => $this->capabilities->isZeroInterest()
                 ];
             }
         }
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -7,3 +7,5 @@
 "Print", "พิมพ์"
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
+"Choose number of monthly payments", "เลือกระยะเวลาผ่อนชำระ"
+"%terms months (%amount THB / month)", "%terms เดือน (%amount บาท/เดือน)"
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -8,4 +8,4 @@
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Choose number of monthly payments", "เลือกระยะเวลาผ่อนชำระ"
-"%terms months (%amount THB / month)", "%terms เดือน (%amount/เดือน)"
+"%terms months (%amount / month)", "%terms เดือน (%amount / เดือน)"
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -9,3 +9,4 @@
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Choose number of monthly payments", "เลือกระยะเวลาผ่อนชำระ"
 "%terms months (%amount / month)", "%terms เดือน (%amount / เดือน)"
+"Manage your cards", "จัดการบัตร"
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -8,4 +8,4 @@
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Choose number of monthly payments", "เลือกระยะเวลาผ่อนชำระ"
-"%terms months (%amount THB / month)", "%terms เดือน (%amount บาท/เดือน)"
+"%terms months (%amount THB / month)", "%terms เดือน (%amount/เดือน)"
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -57,6 +57,37 @@ define(
             },
 
             /**
+             * Get Installment minimum
+             * this function respects info from: https://www.omise.co/installment-payment
+             *
+             * NOTE: in the future this function should return data from capabilities object.
+             *
+             * @return {integer}
+             */
+            getInstallmentMinimum(id) {
+                switch (id) {
+                    case 'kbank': return 500;
+                    case 'bbl': return 500;
+                    case 'bay': return 300;
+                    case 'first_choice': return 300;
+                    case 'ktc': return 300;
+                    default: return NaN;
+                }
+            },
+
+            /**
+             * Checks if minimum amount is respected
+             * 
+             * @return {boolean}
+             */
+            isMinimumAmount(id, terms) {
+                const min = this.getInstallmentMinimum(id);
+                const total = checkoutConfig.quoteData.grand_total;
+
+                return total / terms >= min;
+            },
+
+            /**
              * Get installment terms
              * 
              * @return {string|null}
@@ -122,10 +153,12 @@ define(
 
                         var dispTerms = [];
                         for (let i = 0; i < terms.length; i++) {
-                            dispTerms.push({ 
-                                label: templateLabel.replace('%terms', terms[i]).replace('%amount', (total / terms[i]).toFixed(0)),
-                                key: terms[i] 
-                            });
+                            if (this.isMinimumAmount(id, terms[i])) {
+                                dispTerms.push({
+                                    label: templateLabel.replace('%terms', terms[i]).replace('%amount', (total / terms[i]).toFixed(0)),
+                                    key: terms[i]
+                                });
+                            }
                         }
 
                         return ko.observableArray(
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -60,11 +60,12 @@ define(
 
             /**
              * Format Price
-             *
+             * 
+             * @param {float} amount - Amount to be formatted
              * @return {string}
              */
-            getFormattedPrice: function (price) {
-                return priceUtils.formatPrice(price, quote.getPriceFormat());
+            getFormattedAmount: function (amount) {
+                return priceUtils.formatPrice(amount, quote.getPriceFormat());
             },
 
             /**
@@ -73,15 +74,16 @@ define(
              *
              * NOTE: in the future this function should return data from capabilities object.
              *
+             * @param {string} id - Bank ID
              * @return {integer}
              */
-            getInstallmentMinimum(id) {
+            getInstallmentMinimum: function (id) {
                 switch (id) {
-                    case 'kbank':        return 500;
-                    case 'bbl':          return 500;
-                    case 'bay':          return 300;
+                    case 'kbank': return 500;
+                    case 'bbl': return 500;
+                    case 'bay': return 300;
                     case 'first_choice': return 300;
-                    case 'ktc':          return 300;
+                    case 'ktc': return 300;
                 }
             },
 
@@ -89,16 +91,17 @@ define(
              * Get Installment monthly interest rate
              *
              * NOTE: in the future this function should return data from capabilities object.
-             *
+             * 
+             * @param {string} id - Bank id
              * @return {float}
              */
-            getinstallmentInterestRate(id) {
+            getInstallmentInterestRate: function (id) {
                 switch (id) {
-                    case 'kbank':        return 0.0065;
-                    case 'bbl':          return 0.008;
-                    case 'bay':          return 0.008;
+                    case 'kbank': return 0.0065;
+                    case 'bbl': return 0.008;
+                    case 'bay': return 0.008;
                     case 'first_choice': return 0.013;
-                    case 'ktc':          return 0.008;
+                    case 'ktc': return 0.008;
                 }
             },
 
@@ -109,25 +112,27 @@ define(
              *
              * @return {float}
              */
-            getVatRate() {
+            getVatRate: function () {
                 return 0.07;
             },
 
             /**
              * Get zero interest setting
              *
-             * @return {string}
+             * @return {boolean}
              */
-            isZeroInterest() {
+            isZeroInterest: function () {
                 return window.checkoutConfig.is_zero_interest;
             },
 
             /**
              * Calculates single installment amount
-             *
+             * 
+             * @param {string} id - Bank ID
+             * @param {integer} terms - number of monthly installments
              * @return {integer}
              */
-            calculateSingleInstallmentAmount(id, terms) {
+            calculateSingleInstallmentAmount: function (id, terms) {
                 const total = this.getTotal();
                 const tax = this.getVatRate();
 
@@ -137,22 +142,12 @@ define(
                 }
 
                 // get monthly interest setting
-                const rate = this.getinstallmentInterestRate(id);
+                const rate = this.getInstallmentInterestRate(id);
+
+                //calculate interests
                 const interest = ((1 + tax) * rate * terms * total);
 
-                return + ( ( (total + interest ) / terms ).toFixed( 0 ) );
-            },
-
-            /**
-             * Checks if minimum amount is respected
-             * 
-             * @return {boolean}
-             */
-            isMinimumAmount(id, terms) {
-                const min = this.getInstallmentMinimum(id);
-                const singleInstallment = this.calculateSingleInstallmentAmount(id, terms);
-
-                return singleInstallment >= min;
+                return + (((total + interest) / terms).toFixed(0));
             },
 
             /**
@@ -160,8 +155,8 @@ define(
              *
              * @return {integer}
              */
-            getTotal() {
-                return +window.checkoutConfig.totalsData.grand_total;
+            getTotal: function () {
+                return + window.checkoutConfig.totalsData.grand_total;
             },
 
             /**
@@ -169,14 +164,14 @@ define(
              * 
              * @return {string|null}
              */
-            getTerms() {
+            getTerms: function () {
                 return this.installmentTermsBBL() || this.installmentTermsKBank() || this.installmentTermsFC() || this.installmentTermsKTC() || this.installmentTermsBAY();
             },
 
             /**
              * Reset selected terms
              */
-            resetTerms() {
+            resetTerms: function () {
                 this.installmentTermsBBL(null);
                 this.installmentTermsKBank(null);
                 this.installmentTermsFC(null);
@@ -210,33 +205,36 @@ define(
 
             /**
              * Returns Installment Terms
-             *
+             * @param {string} id - Bank id
              * @return {array}
              */
-            getInstallmentTerms(id) {
+            getInstallmentTerms: function (id) {
                 const installmentBackends = checkoutConfig.installment_backends;
                 const templateLabel = $.mage.__('%terms months (%amount / month)');
 
                 for (const key in installmentBackends) {
-                    if (installmentBackends[key]._id === 'installment_' + id) {
-                        const terms = installmentBackends[key].allowed_installment_terms;
-
-                        var dispTerms = [];
-                        for (let i = 0; i < terms.length; i++) {
-                            if (this.isMinimumAmount(id, terms[i])) {
-                                const amount = this.getFormattedPrice(this.calculateSingleInstallmentAmount(id, terms[i]));
-
-                                dispTerms.push({
-                                    label: templateLabel.replace('%terms', terms[i]).replace('%amount', amount),
-                                    key: terms[i]
-                                });
-                            }
-                        }
-
-                        return ko.observableArray(
-                            dispTerms
-                        );
+                    if (installmentBackends[key]._id !== 'installment_' + id) {
+                        continue;
                     }
+
+                    let dispTerms = [];
+                    const terms = installmentBackends[key].allowed_installment_terms;
+                    const minSingleInstallment = this.getInstallmentMinimum(id);
+
+                    for (let i = 0; i < terms.length; i++) {
+                        const amount = this.calculateSingleInstallmentAmount(id, terms[i]);
+
+                        if (amount >= minSingleInstallment) {
+                            dispTerms.push({
+                                label: templateLabel.replace('%terms', terms[i]).replace('%amount', this.getFormattedAmount(amount)),
+                                key: terms[i]
+                            });
+                        }
+                    }
+
+                    return ko.observableArray(
+                        dispTerms
+                    );
                 }
             },
 
@@ -250,7 +248,7 @@ define(
             },
 
             /**
-             * Get store currency 
+             * Get store currency
              *
              * @return {string}
              */
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -67,6 +67,8 @@ define(
                 };
             },
 
+
+
             /**
              * Is method available to display
              *
@@ -105,6 +107,15 @@ define(
              */
             getOrderCurrency: function () {
                 return window.checkoutConfig.quoteData.quote_currency_code;
+            },
+
+            /**
+             * Get zero interests setting
+             *
+             * @return {string}
+             */
+            isZeroInterests: function () {
+                return window.checkoutConfig.is_zero_interest;
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -66,13 +66,63 @@ define(
              */
             getInstallmentMinimum(id) {
                 switch (id) {
-                    case 'kbank': return 500;
-                    case 'bbl': return 500;
-                    case 'bay': return 300;
+                    case 'kbank':        return 500;
+                    case 'bbl':          return 500;
+                    case 'bay':          return 300;
                     case 'first_choice': return 300;
-                    case 'ktc': return 300;
-                    default: return NaN;
+                    case 'ktc':          return 300;
+                    default:             return NaN;
                 }
+            },
+
+            /**
+             * Get Installment yearly interest rate
+             *
+             * NOTE: in the future this function should return data from capabilities object.
+             *
+             * @return {float}
+             */
+            getinstallmentInterestRate(id) {
+                switch (id) {
+                    case 'kbank':        return 0.08;
+                    case 'bbl':          return 0.08;
+                    case 'bay':          return 0.08;
+                    case 'first_choice': return 0.3;
+                    case 'ktc':          return 0.08;
+                    default:             return NaN;
+                }
+            },
+
+            /**
+             * Get zero interests setting
+             *
+             * @return {string}
+             */
+            isZeroInterests() {
+                return window.checkoutConfig.is_zero_interest;
+            },
+
+            /**
+             * Calculates single installment amount
+             *
+             * @return {integer}
+             */
+            calculateInstallmentAmount(id, terms) {
+                const total = checkoutConfig.quoteData.grand_total;
+
+                if (this.isZeroInterests()) {
+                    return (total / terms).toFixed(0)
+                }
+
+                // get yearly interests setting
+                const rate = this.getinstallmentInterestRate(id);
+
+                // NOTE: total amount should be increased by interests
+                // that customer will pay
+                // increase ratio is calculated using formula:
+                // ratio = 1 + ([yearly interests rate] / [12 months in year] * [number of monthly payments])
+                const incTotal = total * (1 + rate / 12 * terms);
+                return ( incTotal / terms ).toFixed(0);
             },
 
             /**
@@ -144,7 +194,6 @@ define(
              */
             getInstallmentTerms(id) {
                 const installmentBackends = checkoutConfig.installment_backends;
-                const total = checkoutConfig.quoteData.grand_total;
                 const templateLabel = $.mage.__('%terms months (%amount THB / month)');
 
                 for (const key in installmentBackends) {
@@ -154,8 +203,9 @@ define(
                         var dispTerms = [];
                         for (let i = 0; i < terms.length; i++) {
                             if (this.isMinimumAmount(id, terms[i])) {
+                                const amount = this.calculateInstallmentAmount(id, terms[i]);
                                 dispTerms.push({
-                                    label: templateLabel.replace('%terms', terms[i]).replace('%amount', (total / terms[i]).toFixed(0)),
+                                    label: templateLabel.replace('%terms', terms[i]).replace('%amount', amount),
                                     key: terms[i]
                                 });
                             }
@@ -163,7 +213,7 @@ define(
 
                         return ko.observableArray(
                             dispTerms
-                        )
+                        );
                     }
                 }
             },
@@ -175,15 +225,6 @@ define(
              */
             getOrderCurrency: function () {
                 return window.checkoutConfig.quoteData.quote_currency_code;
-            },
-
-            /**
-             * Get zero interests setting
-             *
-             * @return {string}
-             */
-            isZeroInterests: function () {
-                return window.checkoutConfig.is_zero_interest;
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -71,7 +71,6 @@ define(
                     case 'bay':          return 300;
                     case 'first_choice': return 300;
                     case 'ktc':          return 300;
-                    default:             return NaN;
                 }
             },
 
@@ -89,7 +88,6 @@ define(
                     case 'bay':          return 0.008;
                     case 'first_choice': return 0.013;
                     case 'ktc':          return 0.008;
-                    default:             return NaN;
                 }
             },
 
@@ -127,7 +125,7 @@ define(
                     return (total / terms).toFixed(0)
                 }
 
-                // get yearly interests setting
+                // get monthly interests setting
                 const rate = this.getinstallmentInterestRate(id);
                 const interests = ((1 + tax) * rate * terms * total);
 
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -121,10 +121,12 @@ define(
                         const terms = installmentBackends[key].allowed_installment_terms;
                         var dispTerms = [];
                         for (let i = 0; i < terms.length; i++) {
-                            dispTerms.push(`${terms[i]} months (${(total / terms[i]).toFixed(0)} THB / month)`);
+                            dispTerms.push({ label: `${terms[i]} months (${(total / terms[i]).toFixed(0)} THB / month)`, key: terms[i]});
                         }
 
-                        return dispTerms;
+                        return ko.observableArray(
+                            dispTerms
+                        )
                     }
                 }
             },
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -7,7 +7,8 @@ define(
         'Magento_Checkout/js/model/full-screen-loader',
         'Magento_Checkout/js/model/quote',
         'Magento_Checkout/js/model/url-builder',
-        'Magento_Checkout/js/model/error-processor'
+        'Magento_Checkout/js/model/error-processor',
+        'Magento_Catalog/js/price-utils',
     ],
     function (
         $,
@@ -17,7 +18,8 @@ define(
         fullScreenLoader,
         quote,
         urlBuilder,
-        errorProcessor
+        errorProcessor,
+        priceUtils,
     ) {
         'use strict';
 
@@ -54,6 +56,15 @@ define(
                     ]);
 
                 return this;
+            },
+
+            /**
+             * Format Price
+             *
+             * @return {string}
+             */
+            getFormattedPrice: function (price) {
+                return priceUtils.formatPrice(price, quote.getPriceFormat());
             },
 
             /**
@@ -204,7 +215,7 @@ define(
              */
             getInstallmentTerms(id) {
                 const installmentBackends = checkoutConfig.installment_backends;
-                const templateLabel = $.mage.__('%terms months (%amount THB / month)');
+                const templateLabel = $.mage.__('%terms months (%amount / month)');
 
                 for (const key in installmentBackends) {
                     if (installmentBackends[key]._id === 'installment_' + id) {
@@ -213,7 +224,7 @@ define(
                         var dispTerms = [];
                         for (let i = 0; i < terms.length; i++) {
                             if (this.isMinimumAmount(id, terms[i])) {
-                                const amount = this.calculateSingleInstallmentAmount(id, terms[i]);
+                                const amount = this.getFormattedPrice(this.calculateSingleInstallmentAmount(id, terms[i]));
 
                                 dispTerms.push({
                                     label: templateLabel.replace('%terms', terms[i]).replace('%amount', amount),
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -172,12 +172,6 @@ define(
                 };
             },
 
-            getFormattedInstallmentTerms: function () {
-                var termsStr = getTerms();
-                console.log(termsStr);
-                return termsStr.substring(0, termsStr.indexOf('months') - 1);
-            },
-
             /**
              * Is method available to display
              *
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -106,17 +106,6 @@ define(
             },
 
             /**
-             * Get VAT tax rate
-             *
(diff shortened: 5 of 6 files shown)


view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -137,7 +255,7 @@ define(
              * @return {boolean}
              */
             orderValueTooLow: function () {
-                return window.checkoutConfig.totalsData.base_grand_total < 3000;
+                return this.getTotal() < 3000;
```
guzzilar: @jacstn regarding to the Magento's implementation and the way they use `base_total`, `total`, `base_grand_total`, `grand_total`. Having a method that change the definition of `base_grand_total` to `getTotal` may lead to a misunderstanding?

jacstn: i think grand total is better than base grand total according to this article:
https://stackoverflow.com/questions/17999330/what-is-the-difference-among-base-grand-total-grand-total-sub-total-in-magent

because grand total is always displaying in shop main currency, where base_grand_total keeps value in current chosen currency (if user defined one)

jacstn: i return there `grand_total`, changed from `base_grand_total`

Model/Api/Capabilities.php
```diff
@@ -24,4 +24,13 @@ public function getInstallmentBackends()
             $this->capabilities->makeBackendFilterType('installment')
         );    
     }
+
+    /**
+     * Get information about zero interest installments
+     *
+     * @return bool
+     */
+    public function isZeroInterests() {
```
jonrandy: Should be `isZeroInterest`

Model/Capabilities.php
```diff
@@ -36,4 +36,12 @@ public function retrieveInstallmentBackends()
     {
         return $this->capabilitiesAPI->getInstallmentBackends();
     }
+
+    /**
+     * @return bool
+     */
+    public function isZeroInterests()
```
jonrandy: and again

i18n/th_TH.csv
```diff
@@ -7,3 +7,5 @@
 "Print", "พิมพ์"
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
+"Choose number of monthly payments", "เลือกระยะเวลาผ่อนชำระ"
+"%terms months (%amount THB / month)", "%terms เดือน (%amount บาท/เดือน)"
```
jonrandy: Should we be using Magento's formatting function for a currency amount?

jacstn: we don't use magento function for for formatting.

jacstn: i got that from framework.. followed what they did in magento.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+             * @return {integer}
+             */
+            getInstallmentMinimum(id) {
+                switch (id) {
+                    case 'kbank':        return 500;
+                    case 'bbl':          return 500;
+                    case 'bay':          return 300;
+                    case 'first_choice': return 300;
+                    case 'ktc':          return 300;
+                    default:             return NaN;
```
jonrandy: Not sure why defaulting to `NaN`? Won't this break stuff if a new instalment provider appears? Or are they hardcoded?

jacstn: no new installment provider until we hardcode one.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+             * @return {float}
+             */
+            getinstallmentInterestRate(id) {
+                switch (id) {
+                    case 'kbank':        return 0.0065;
+                    case 'bbl':          return 0.008;
+                    case 'bay':          return 0.008;
+                    case 'first_choice': return 0.013;
+                    case 'ktc':          return 0.008;
+                    default:             return NaN;
```
jonrandy: Same as above

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: And if they are hardcoded, this `default` is redundant?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+            getVatRate() {
+                return 0.07;
+            },
+
+            /**
+             * Get zero interests setting
+             *
+             * @return {string}
+             */
+            isZeroInterests() {
```
jonrandy: `isZeroInterest`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: it is fixed, this default is to check when testing if everything works as expected, and data is displayed, otherwise would be NaN.
But actually that `default` will never affect anything as we hardcode stuff in `.html`
so if new provider will show up it will return NaN, but it will not be displayed anywhere in UI

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+                if (this.isZeroInterests()) {
+                    //merchant pays interests
+                    return (total / terms).toFixed(0)
+                }
+
+                // get yearly interests setting
+                const rate = this.getinstallmentInterestRate(id);
+                const interests = ((1 + tax) * rate * terms * total);
+
+                return + ( ( (total + interests ) / terms ).toFixed( 0 ) );
```
jonrandy: What is `+` doing there?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+            calculateSingleInstallmentAmount(id, terms) {
+                const total = this.getTotal();
+                const tax = this.getVatRate();
+
+                if (this.isZeroInterests()) {
+                    //merchant pays interests
+                    return (total / terms).toFixed(0)
+                }
+
+                // get yearly interests setting
```
jonrandy: Yearly?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,112 @@ define(
...
+                return singleInstallment >= min;
+            },
+
+            /**
+             * Get total amount of an order
+             *
+             * @return {integer}
+             */
+            getTotal() {
+                return +window.checkoutConfig.totalsData.grand_total;
```
jonrandy: Why the `+`?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: it is converting string to integer

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Shortcut to convert from string?

jacstn: the same, conversion.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -107,9 +206,28 @@ define(
              */
             getInstallmentTerms(id) {
                 const installmentBackends = checkoutConfig.installment_backends;
+                const templateLabel = $.mage.__('%terms months (%amount THB / month)');
```
jonrandy: Should we be following the standard currency amount formatting for magento?

jacstn: example please..

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -107,9 +206,28 @@ define(
...
+
                 for (const key in installmentBackends) {
-                    if (installmentBackends[key]._id === 'installment_' + id)
-                        return ko.observableArray(installmentBackends[key].allowed_installment_terms);
+                    if (installmentBackends[key]._id === 'installment_' + id) {
+                        const terms = installmentBackends[key].allowed_installment_terms;
+
+                        var dispTerms = [];
+                        for (let i = 0; i < terms.length; i++) {
```
jonrandy: Simpler to use `forEach`?

jacstn: well maybe.. this is a habit from `nodejs`.. `foreach` takes way more resources than `for`..

view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -64,7 +64,12 @@
                             </div>
                         </label>
                         <select style="margin-top:10px" data-bind="options: getInstallmentTerms('ktc'),
```
jonrandy: Would this be better in CSS? This hardcoded styling may break someone's theme?

jacstn: this is some past change outside of this PR. Will add it to todo list for future pr

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: I don't know how this is handled in Magento 2. Magento 1 certainly has functions for this

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: ok

i18n/th_TH.csv
jonrandy: Fair enough. It's handled by magento functions in Mag1

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: So the `default` can be removed

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: well.. I don't format currency here, it is JS.
I display just simply amount without decimals as number minimum is 300.

jacstn: what od you use in Mag 1?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: default is good practice.. that is how they taught me during studies..
it is that you have covered everything.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: `Mage::app()->getLocale()->currency($currencyCode)->toCurrency($amount)`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: but might delete that.. if omise has different rules.. @guzzilar

jonrandy:
```js
getInstallmentMinimum(id) {
   return {kbank:500, bbl:500}[id] || 300;
}
```
Would be another way to go

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: for now, those limits are fixed, pirces are counted by fixed results, it has to be thai baht. getting that from magento, would bring confusion or wrong data.
if user will switch price to display it ie in GBP, this will still show data according to THB. we need to force users to see THB here. (not sure if i described that understandebly)

jacstn: however, found in mag 2 from js that there are simillar mechanisms to mag 1 to display currencies

Model/Api/Capabilities.php
jonrandy: Unfixed?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: ok. Worth using it?

jacstn: not in this case.

jacstn: it must be THB always..

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,110 @@ define(
...
+            /**
+             * Calculates single installment amount
+             *
+             * @return {integer}
+             */
+            calculateSingleInstallmentAmount(id, terms) {
+                const total = this.getTotal();
+                const tax = this.getVatRate();
+
+                if (this.isZeroInterest) {
```
guzzilar: @jacstn  no nee to execute this one? `this.isZeroInterest()`

jacstn: fixed.

guzzilar:
IIRC the nature of Magento for the `base_grand_total` and `grand_total`, `grand_total` would be used to display on a screen only but when it comes to payment processing, it always use `base_grand_total`. I want to check on this topic a bit more.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -137,7 +253,7 @@ define(
              * @return {boolean}
              */
             orderValueTooLow: function () {
-                return window.checkoutConfig.totalsData.base_grand_total < 3000;
+                return this.getTotal() < 3000;
```
guzzilar: @jacstn wonder what is the idea behind this modification, why should we have an another function (`getTotal()`) to encapsulate this `window.checkoutConfig.totalsData.base_grand_total`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: But then this amount won't be accurated right?
I.e. A store base currency is `THB` and the display currency I chose is `USD`.
When I pick an item that its amount is 100 USD at `grand_total`, (meaning, about 3,1xx THB at `base_grand_total`) but this store will show me that my purchase amount is too low and I have to pick more item to make its amount exceed 3,000 USD (at `grad_total`) which way far more exceed than 3,000 THB at `base_grand_total`.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,110 @@ define(
...
+            /**
+             * Get Installment minimum
+             * this function respects info from: https://www.omise.co/installment-payment
+             *
+             * NOTE: in the future this function should return data from capabilities object.
+             *
+             * @return {integer}
+             */
+            getInstallmentMinimum(id) {
+                switch (id) {
```
guzzilar: An array or an object solution would be more efficient? Don't see much point to use Switch condition here.
(IIRC, saw @jonrandy commented once?)

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -38,7 +38,8 @@ public function getConfig()
         foreach ($listOfActivePaymentMethods as $method) {
             if ($method->getCode() === OmiseInstallmentConfig::CODE) {
                 return [ 
-                    'installment_backends' => $this->capabilities->retrieveInstallmentBackends()
+                    'installment_backends' => $this->capabilities->retrieveInstallmentBackends(),
+                    'is_zero_interest' => $this->capabilities->isZeroInterest()
```
guzzilar: Saw that we have to do such tasks to manipulate the list of instalment data at JS file. 
Wouldn't it be better to finish those tasks here before returning it out to frontend? Wonder what is the idea behind this approach (to manipulate list data at the client side).

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: yeah.. maybe, here is more readable IMHO.

jacstn: @guzzilar did you test it previous `grand_total` comment? we force THB, no THB installment will be not selectable. But might be some my mistake here..

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: we use it more than once. Was thinking that it is more readable than this long structure,

also we have stuff like:

```
            /**
             * Get Omise public key
             *
             * @return {string}
             */
            getPublicKey: function() {
                return window.checkoutConfig.payment.omise_cc.publicKey;
            },```

File: view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js (hunk 11 above)
Comment (guzzilar) [reply view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js]: In that case, I think we better use the same Magento format?
Screen Shot 2562-05-24 at 22 07 17

`฿36,999.94 / month`
Or even better if we can use a formatter function provided by Magento, cuz there is a chance that merchant modified their store in a different format from default (i.e. some store uses `฿36,999.94 Baht / month`, some uses `36,999.94 Baht / month`.

These options are protected by `isActive()` function anyway so it shouldn't be any confusion?

File: view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js (hunk 16 above)
Comment (guzzilar) [reply view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js]: Readable, probably. But feel a bit off when it comes to code intention.
If you want to remain its readability then why not just:
```js
return {
  'kbank': 500,
  'bbl': 500,
  'bay': 300,
  'first_choice': 300,
  'ktc': 300
}[id] || 0;
```
Or could even be
```js
const minAllowedAmount = {
  'kbank': 500,
  'bbl': 500,
  'bay': 300,
  'first_choice': 300,
  'ktc': 300
}

return minAllowedAmount[id] || 0;
```
(and yes, I agreed on having a fallback condition)

--
A little add, I think `id` is a bit ambiguous in this case... would prefer `bank_code`, a little long but more meaningful.
```js
return minAllowedAmount[bank_code] || 0;
```
Anyway, just personal preference, I'll leave it to you.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: Just in case, the translation part could be `"%terms months (%amount / month)", "%terms เดือน (%amount / เดือน)"`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: For the question
> @guzzilar did you test it previous grand_total comment? we force THB, no THB installment will be not selectable. But might be some my mistake here..

Let me move to this thread.

My fault, yup, it's protected by `isActive()` function already.
However, one little note.
Screen Shot 2562-05-24 at 22 00 41

It would be nice if we can hide that `order value too low` message if the method isn't active (actually, how about `order amount` instead of `order value`).

Maybe `visible: (isActive() && orderValueTooLow())` could help?
Anyway, feel free to click resolve this conversation if you don't have anything to add. (about the warning message, I think we can do it later as it is out of the scope of this PR)

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,13 +56,110 @@ define(
...
+                if (this.isZeroInterest()) {
+                    //merchant pays interest
+                    return (total / terms).toFixed(0)
+                }
+
+                // get monthly interest setting
+                const rate = this.getinstallmentInterestRate(id);
+                const interest = ((1 + tax) * rate * terms * total);
+
+                return + ( ( (total + interest ) / terms ).toFixed( 0 ) );
```
guzzilar: @jacstn this one I forgot to test, but wonder why we want to round up those decimal? Why not leave it as `toFixed(2)` or there is no case that this would be 2 digits decimal?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: ic

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: we round up because the minimum is 300 THB, so in case of such a big number showing decimals does not matter (I think), it is only to show people how much more or less they will pay, the rest they will see on bank pages.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: The formatting of THB amounts doesn't seem to be following Magento settings

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -107,9 +204,28 @@ define(
...
                 const installmentBackends = checkoutConfig.installment_backends;
+                const templateLabel = $.mage.__('%terms months (%amount THB / month)');
+
                 for (const key in installmentBackends) {
-                    if (installmentBackends[key]._id === 'installment_' + id)
-                        return ko.observableArray(installmentBackends[key].allowed_installment_terms);
+                    if (installmentBackends[key]._id === 'installment_' + id) {
+                        const terms = installmentBackends[key].allowed_installment_terms;
+
+                        var dispTerms = [];
```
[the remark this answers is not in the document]
jacstn: @guzzilar end result is the same, there is the only a small difference with memory management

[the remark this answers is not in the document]
jacstn: i have been always using `disp` as display
https://acronyms.thefreedictionary.com/DISP

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -107,9 +204,28 @@ define(
              */
             getInstallmentTerms(id) {
                 const installmentBackends = checkoutConfig.installment_backends;
+                const templateLabel = $.mage.__('%terms months (%amount THB / month)');
+
                 for (const key in installmentBackends) {
-                    if (installmentBackends[key]._id === 'installment_' + id)
-                        return ko.observableArray(installmentBackends[key].allowed_installment_terms);
+                    if (installmentBackends[key]._id === 'installment_' + id) {
```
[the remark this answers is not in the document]
jacstn: yes.. that could be done

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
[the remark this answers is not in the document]
guzzilar: @jacstn ok, should we be consistent and go with one style?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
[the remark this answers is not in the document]
guzzilar: then let's do it :D

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -107,9 +204,28 @@ define(
...
+                        const terms = installmentBackends[key].allowed_installment_terms;
+
+                        var dispTerms = [];
+                        for (let i = 0; i < terms.length; i++) {
+                            if (this.isMinimumAmount(id, terms[i])) {
```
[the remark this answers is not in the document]
jacstn: that is very good point @guzzilar

Model/Ui/CapabilitiesConfigProvider.php
guzzilar: That's interesting. Do you mind to share me some resource or article that explains about a benefit of moving the data manipulation from server side to client side? Can be some later time, just want to study on this.

mark as resolved for now.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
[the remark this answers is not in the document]
guzzilar: At first glance, I was reading it as `dispute terms`.
Not a big fan of abbreviating a variable name unless it's in a local scope (like `for ($i =0; ...){}`). Can you consider not to abbreviate it and just go as `displayTerms`? Just 3 letters longer, but eliminate any potential of misinterpretation.

Gonna mark as resolved and leave the decision to you.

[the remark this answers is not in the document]
jacstn: no, it does not matter.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: i translated it following info from Carol. Is that not correct?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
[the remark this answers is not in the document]
guzzilar: Code consistency is not matter? I doubt... 🤔

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: The translation was corrected. Just I'm suggesting the idea that we can remove `THB` out from the translation and leave it to Magento's formatter function to handle it.
As currency format can be various depending on each store.
- `฿36,999.94 / month`
- `฿36,999.94 Baht / month`
- `36,999.94 Baht / month`


Original: `"%terms months (%amount THB / month)", "%terms เดือน (%amount บาท/เดือน)"`
If we format the currency code by using Magento formatter then those `%amount THB / month`, `%amount บาท/เดือน` can become `%amount / month`, `%amount / เดือน`.

i18n/th_TH.csv
```diff
@@ -7,4 +7,5 @@
 "Print", "พิมพ์"
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
-"Manage your cards", "จัดการบัตร"
```
guzzilar: @jacstn maybe accidentally deleted?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+             */
+            getInstallmentMinimum: function (id) {
+                switch (id) {
+                    case 'kbank': return 500;
+                    case 'bbl': return 500;
+                    case 'bay': return 300;
+                    case 'first_choice': return 300;
+                    case 'ktc': return 300;
+                }
+            },
```
guzzilar: Screen Shot 2562-05-28 at 05 52 06

Not to bring this topic back, but if you are talking about its readability, I think the right screen is way more cleaner 😂 
(just share opinion, not asking for change)

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+                const total = this.getTotal();
+                const tax = this.getVatRate();
+
+                if (this.isZeroInterest()) {
+                    //merchant pays interest
+                    return (total / terms).toFixed(0)
+                }
+
+                // get monthly interest setting
+                const rate = this.getInstallmentInterestRate(id);
```
guzzilar: IMO, if a method or a function has been named properly, we wouldn't need to have a code comment here.
If we feel to comment it, maybe we should consider to rename the function.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+             * @param {integer} terms - number of monthly installments
+             * @return {integer}
+             */
+            calculateSingleInstallmentAmount: function (id, terms) {
+                const total = this.getTotal();
+                const tax = this.getVatRate();
+
+                if (this.isZeroInterest()) {
+                    //merchant pays interest
+                    return (total / terms).toFixed(0)
```
guzzilar: same as the below comment, if we feel a need to have a code comment here, then maybe we should consider to rename `this.isZeroInterest()` to `this.isMerchantPaysInterest()`.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+                }
+
+                // get monthly interest setting
+                const rate = this.getInstallmentInterestRate(id);
+
+                //calculate interests
+                const interest = ((1 + tax) * rate * terms * total);
+
+                return + (((total + interest) / terms).toFixed(0));
```
guzzilar: About `toFixed(0)`, can you help confirm with the team if it's ok to round up those decimals? I'm not sure if it's the best way to cut all decimals out.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -123,7 +248,7 @@ define(
             },
 
             /**
-             * Get store currency
+             * Get store currency
```
guzzilar: accidentally adding space?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -102,14 +205,36 @@ define(
...
+                    for (let i = 0; i < terms.length; i++) {
+                        const amount = this.calculateSingleInstallmentAmount(id, terms[i]);
+
+                        if (amount >= minSingleInstallment) {
+                            dispTerms.push({
+                                label: templateLabel.replace('%terms', terms[i]).replace('%amount', this.getFormattedAmount(amount)),
+                                key: terms[i]
+                            });
+                        }
+                    }
```
guzzilar: What if we add `break;` here to preventing an unnecessary iteration?
(i.e. if term 8 months is lower than `minSingleInstallment` then, stop it from calculating for 9 months, 10 months, 12 months, and so on)
```js
for (let i = 0; i < terms.length; i++) {
    const amount = this.calculateSingleInstallmentAmount(id, terms[i]);

    if (amount < minSingleInstallment) {
        break;
    }

    dispTerms.push({
        label: templateLabel.replace('%terms', terms[i]).replace('%amount', this.getFormattedAmount(amount)),
        key: terms[i]
    });
}
```

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+             * @return {float}
+             */
+            getVatRate: function () {
+                return 0.07;
+            },
+
+            /**
+             * Get zero interest setting
+             *
+             * @return {string}
```
guzzilar: return `{boolean}` maybe?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: what guarantee we have that it will be this way sorted every time when coming from backend?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: I am fine with the current solution.

jonrandy:
In the screenshot, the monthly amounts appear to be rounded to the nearest baht? This isn't right, surely?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: I prefer the object based solution too

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+             * Get Installment monthly interest rate
+             *
+             * NOTE: in the future this function should return data from capabilities object.
+             * 
+             * @param {string} id - Bank id
+             * @return {float}
+             */
+            getInstallmentInterestRate: function (id) {
+                switch (id) {
+                    case 'kbank': return 0.0065;
```
jonrandy: Feels wrong to have these hardcoded in JS - wouldn't a better place be a config file?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Also feels wrong to have these hardcoded in JS. In a config file would be better?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -56,19 +58,120 @@ define(
...
+                    case 'bay': return 0.008;
+                    case 'first_choice': return 0.013;
+                    case 'ktc': return 0.008;
+                }
+            },
+
+            /**
+             * Get VAT tax rate
+             *
+             * NOTE: in the future this function should return data from capabilities object.
```
jonrandy: Really? Doesn't seem related to the capabilities of a merchant account? Rather the economic policy of a country. Should come from config?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -137,7 +262,7 @@ define(
              * @return {boolean}
              */
             orderValueTooLow: function () {
-                return window.checkoutConfig.totalsData.base_grand_total < 3000;
```
jonrandy: Magic numbers. Should be in config?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: no, as heder of function says, it will come from capabilities object, it is all available here..

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: it will come from capabilities api, entire object is available here in js.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: we need to have it from capabilities. if vat will change for some reason is only one click on backend. in poland vat has been changed few times for last years.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: this will come from capability api. that 3000 thb

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: These are not in the new capabilities API, and there is no plan to put them there AFAIK

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: These are not in the new capabilities API, and there is no plan to put them there AFAIK. VAT rate is not related to merchant account capabilities

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: These are not in the new capabilities API, and there is no plan to put them there AFAIK

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: it needs to come from capabilities it is a matter of discussion with backend team.

changing plugins when those conditions will change it is huge pain.

also those companies have some promotions, that we will have easy to manage when it will come from capabilities.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: depends how you look at it. vat in thailand is for particular merchant his capability, he has different VAT then in Japan.

jonrandy:
Doesn't alter the fact it's nothing to do with the capabilities of the
merchant account

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: in my opinion it is.. vat in different countries is different.
and vat can be changed.
imagine Thailand is increasing vat to 8 % like poland did many times.. and we have to rebuilt all plugins? that is crazy! it needs to be in capabilities.

when they will be developed more and more stuff will be added, and it depends on us what we will want and push backend team to do.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: the same as above.. since even we were developing this functionality minimum has been changed from 5k to 3k.. there is chance it will be changed again. 
It needs to be in capabilities.

jonrandy:
It should ideally be on the Omise servers for sure, but not in capabilities
API (I'm talking about VAT). It certainly doesn't belong there. Until it's
decided where it will come from, the best place for now is somewhere inside
the plugin, with some function that retrieves it. Then when/if it moves
into the API we just need change one function. Hardcoding it right in the
JS seems very bad practice.

I understand that a lot of stuff is done client side in Magento 2, but
surely there's a simple way that a bunch of bars you may want access to can
be passed to the page when it loads client side?

As for the minimum amounts etc. we've already been told these won't be in
the capability API. They aren't in the new capability API and Robin
specifically told me that they won't be

jonrandy:
*vars not bars

jacstn:
so you say that hardcoding in php is better practice than hardcoding in js. why?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: VAT removed, no problem anymore..

Model/Api/Capabilities.php
jonrandy: Did you miss this?

Model/Capabilities.php
jonrandy: Missed?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Missed this?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Missed this?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Missed this?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Is this in Baht? Or currency subunits? We shouldn't be rounding to nearest baht if so, should be nearest satang

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Should also be 'interest'

jonrandy: Also, var name below should be `interest`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: Again, should we be rounding to baht? (if this is baht and not subunits?)

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jonrandy: This currency formatting issue is still not resolved?

i18n/th_TH.csv
```diff
@@ -7,4 +7,6 @@
 "Print", "พิมพ์"
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
```
jonrandy: Currency formatting. Needs to be consistent with other bits that were fixed

Model/Api/Capabilities.php
jacstn: that has been fixed some time ago..

Model/Capabilities.php
jacstn: that has been fixed some time ago @jonrandy I think you look at outdated stuff as even here.. is written
image

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: you are looking at outdated code @jonrandy

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: this is outdated code @jonrandy

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: again.. outdated code @jonrandy

i18n/th_TH.csv
jacstn: that is unrelated to current pr, but I have it in mind.. everything will be changed to the magento currency formatting @jonrandy

jonrandy: Given that the other currency formatting issues have been fixed in this PR, I would suggest that this is now directly related to this PR since not fixing it now results in a feature with inconsistent currency formatting. 

Actually, searching the repo for this string shows it is only in the translation file and nowhere else. Is it even still used?

jacstn: well.. when user make shopping and see that installment terms he see currency formatting in the same way as he sees currency formatting displayed on summary up on the right side of that installment option. it is ok.

that 3k thb not formatted value is visible when the user makes shopping and the minimum requirement is not met. Both will never be displayed from our plugin at the same time.

Also, I wish to remind that it is PR, it is not release, single PR solves the single issue, we will add stuff when we will be doing a release.

jacstn: _Actually, searching the repo for this string shows it is only in the translation file and nowhere else. Is it even still used?_

Some merge errors... i did merge from master ... this string exists in master branch in html, not sure why html was not merged here.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: changed to 2 decimals @guzzilar 
i was big fan of not showing decimals but now magento framework is showing those decimals, so it would show XXXXXXX.00 - that would be ridiculous.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: no more fixed to 0 @guzzilar

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: i removed that comments @guzzilar

i18n/th_TH.csv
jacstn: reverted @guzzilar

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: mater of taste.. IMHO code is more readable if I show all installment terms of all brands.
they also might be changed and must come from capabilities API.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: I changed formatting @jonrandy in next PR #202
@guzzilar also your condition is added there with showing that message

i18n/th_TH.csv
jacstn: @jonrandy I improved that message in #202

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: Thanks @jacstn

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: @jacstn Thanks for re-opened this comment again. I'll mark it as resolved now 👍 👍

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
guzzilar: 👍 Awesome! Resolved :D

PR #197, opened 2019-05-21
Manage cards link on checkout page

jacstn:
#### 1. Objective

This PR adds a link to checkout / payments / credit cards page, to easily go to managing cards page where a user can delete cards.

#### 2. Description of change

Just adds link and TH translation

#### 3. Quality assurance

- **PHP version**: 7.0.16.

**✏️ Details:**

1. Make sure you have saved any credit cards.
Go to checkout, choose credit card payment, see if *Manage your cards* link is available. Click on the link, and see if you are moved to customer panel/managing cards page.

image

2. Make sure you don't have any saved cards.
*Manage your cards* link should not be visible.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -7,3 +7,4 @@
 "Print", "พิมพ์"
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
+"Manage your cards", "จัดการบัตร"
--- a/view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
@@ -62,17 +62,6 @@ define(
             },
 
             /**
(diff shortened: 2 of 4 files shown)


jonrandy:
"Select a card you want to proceed" doesn't make sense - "Select a card to proceed" is better

jacstn:
for me @guzzilar it is easy to imagine when you come back to the store after some time and you see some old cards saved... no need to search in platform how to manage them, easy to imagine that it will be quite useful.

jacstn:
but `target="_blank"` could be better here.

guzzilar:
@jacstn  Still don't see much point in a real use case here. 
Maybe we feel like we need a link to be there because we mostly "test" the system so there will be a lot of cases where we feel the need to jump between "deleting cards" and "place an order".

But in reality, will you just come to the checkout page just to delete a saved card?
Or will user really going to intervene themselves from finish the purchase to delete a card first?

jacstn:
you will not go specifically to the checkout page to delete saved card, it is just option.

PR #198, opened 2019-05-23
Display info on checkout if plugin is in Sandbox mode

jacstn:
#### 1. Objective

This PR adds information on the checkout page if the omise plugin is in sandbox mode. 

#### 2. Description of change

Added new config file that sends information to javascript renderers if the omise plugin is in sandbox mode.

image

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

To manually test this PR:
1. Enable omise plugin sandbox mode in Admin panel.
Check if the warning about the plugin being in sandbox mode is displayed for each of omise payment methods.

2. Disable omise plugin sandbox mode in Admin panel.
Check if none sandbox warning is displayed.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -8,5 +8,5 @@
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Manage your cards", "จัดการบัตร"
-"more info", "โ้อมูลเพิ่มเติม"
+"more info", "ข้อมูลเพิ่มเติม"
 "Test mode", "โหมดทดสอบ"
--- a/i18n/th_TH.csv
+++ b/i18n/th_TH.csv
@@ -9,4 +9,4 @@
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Manage your cards", "จัดการบัตร"
 "more info", "โ้อมูลเพิ่มเติม"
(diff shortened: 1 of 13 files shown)


jacstn:
yes and no... matter of own feeling. It is only warning to the merchant, yes. He should switch to Live asap

but if he will not and customers will come to make shopping than he can see what other payment methods he should use and what definitely not.

But I agree with both solutions, your and mine.
It is just different way to show warning covering different scenarios, where the goal of that warnings is only to make merchant use live mode asap.

jonrandy:
Should probably run it past whoever fielded the complaints from the
merchants

i18n/th_TH.csv
```diff
@@ -8,3 +8,5 @@
 "Choose your installment terms","เลือกระยะเวลาการผ่อนชำระ"
 "Miminum order value is 3000 THB", "ยอดสั่งซื้อขั้นต่ำ 3000 บาท"
 "Manage your cards", "จัดการบัตร"
+"more info", "โ้อมูลเพิ่มเติม"
```
guzzilar: besides of the design, this should be `ข้อมูลเพิ่มเติม`

i18n/th_TH.csv
```diff
@@ -8,3 +8,5 @@
...
+"Test mode", "โหมดทดลอง"
```
guzzilar: maybe `โหมดทดสอบ` is better (saw Omise document uses `โหมดทดสอบ` instead of `โหมดทดลอง`)

PR #199, opened 2019-05-27
Add a very visible 'Omise test mode' reminder on the checkout page 

jonrandy:
#### 1. Objective

Merchants have requested that there be some visual indication on the checkout page if the Omise payment plugin is set to test mode. This will avoid the situation where they may accidentally leave their store in this test payments state

#### 2. Description of change

A small JS file is injected into the checkout page if the config setting for 'test mode' is switched on. This JS adds some HTML to the page to show a warning overlaid at the top of the page. The plugin CSS has also been altered to include styling for the warning.

image

#### 3. Quality assurance

Switch test mode on/off via the plugin settings, make sure the warning on the checkout page appears/disappears as expected.

#### 4. Impact of the change

Merchants way less likely to accidentally leave Omise payments in test mode on their sites

#### 5. Priority of change

Normal

#### 6. Additional Notes

None

### DIFF
--- a/src/skin/frontend/base/default/omise/gateway/css/style.css
+++ b/src/skin/frontend/base/default/omise/gateway/css/style.css
@@ -97,3 +97,19 @@
 .banks-list .omise-logo-wrapper > div.logo-ktc {
     background: url('../images/ktc.svg') #ffffff;
 }
+
+.omise-test-warning {
+    background: #ffcc00;
+    color: #000;
+    font-weight: bold;
+    width: 80%;
+    position: fixed;
+    top: 0;
+    left: 10%;
+    text-transform: uppercase;
+    text-align: center;
+    padding: 5px;
+    border-radius: 0 0 10px 10px;
+    border-top: none;
+    opacity: 0.75;
+}
--- a/src/skin/frontend/base/default/omise/gateway/css/style.css
+++ b/src/skin/frontend/base/default/omise/gateway/css/style.css
@@ -106,10 +106,13 @@
     position: fixed;
     top: 0;
     left: 10%;
-    text-transform: uppercase;
     text-align: center;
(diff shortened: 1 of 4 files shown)


src/skin/frontend/base/default/omise/gateway/js/testmode.js
```diff
@@ -0,0 +1,8 @@
+window.addEventListener('load', function() {
+	var warning = document.createElement('div'); 
+	warning.innerHTML = Translator.translate('OMISE TEST MODE') + ' - \
+		<a target="_blank" href="https://www.omise.co/what-are-test-keys-and-live-keys">\
+		[ ' + Translator.translate('more info') +' ]</a>';
+	warning.className = 'omise-test-warning';
+	document.body.appendChild(warning);
+});
```
jacstn: small thing.. we could add break line here.

jacstn: @jonrandy you are rewritting html with no respect to any module from Magento that is designed for that. are you sure those HTML/CSS classes will exists in every themes?

src/skin/frontend/base/default/omise/gateway/css/style.css
```diff
@@ -97,3 +97,22 @@
 .banks-list .omise-logo-wrapper > div.logo-ktc {
     background: url('../images/ktc.svg') #ffffff;
 }
+
+.omise-test-warning {
```
jacstn: I would use existing class.. so if somebody has some theme installed it will look nice...

jonrandy: I disagree. It needs to look as alien and visible as possible, so they won't miss it. Looking nice is not required

PR #201, opened 2019-05-30
True Money Payment Method Implementation

jacstn:
#### 1. Objective
Enable True Money Payment support.
More info about TrueMoney payment available here: https://www.omise.co/truemoney

#### 2. Description of change

Added extra element to fulfill condition in a callback.
`true money` calls final callback twice. The first time it is sent `payment pending` the second time it is `payment processing`

Added Virtual and PHP classes to support True Money payment.
Added `javascript` and `HTML` UI rendering files.

UI is very simple, and flow from a user perspective is the same as in the case of Alipay.
No extra options are shown to the user.

Checkout page:
image

TrueWallet page:

image

Checkout success page:
image

Admin Panel:
image

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 2.2.7.
- **Omise plugin version**: Omise-Magento 2.2.4.
- **PHP version**: 7.0.29.

**✏️ Details:**

Turn on a `true money` in the admin panel, check if payment is available on the checkout page.
Turn off a `true money` payment method on admin panel, the option should not be listed on checkout page anymore.
Make an order and pay with a `true money` payment method.

#### 4. Impact of the change

none

#### 5. Priority of change

Normal

#### 6. Additional Notes

none


Controller/Callback/Offsite.php
```diff
@@ -66,7 +67,7 @@ public function execute()
             return $this->redirect(self::PATH_CART);
         }
 
-        if ($order->getState() !== Order::STATE_PENDING_PAYMENT) {
+        if (! in_array($order->getState(), [ Order::STATE_PENDING_PAYMENT, Order::STATE_PROCESSING ])) {
```
jacstn: @guzzilar do you have any idea.. why it could send me here PROCESSING, not PENDING like all other offsite payment method? it is Callback.
:(

jacstn: both alipay, and truemoney returns status from omise gateway `"status": "pending"` (found inside response json)
but for some reason, the mag 2 framework returns in `truemoney` case status `processing`, and `alipay` - `pending`

jacstn: this callback offsite is called twice by truemoney.

jacstn:
```[2019-06-11 18:25:16] main.DEBUG: offsite callback execute {"state":"pending_payment"} []
[2019-06-11 18:25:18] main.DEBUG: offsite callback execute {"state":"processing"} []```

Comment (jonrandy) [reply Controller/Callback/Offsite.php]: So, the second one comes as a request from Omise servers to the Magento site? Rather than a redirection to the callback URL?

File: Controller/Callback/Offsite.php [hunk 2]
```diff
@@ -12,6 +12,7 @@
 use Omise\Payment\Model\Config\Internetbanking;
 use Omise\Payment\Model\Config\Alipay;
 use Omise\Payment\Model\Config\Installment;
+use Omise\Payment\Model\Config\Truemoney;
 
 class Offsite extends Action
```
jonrandy: After reading the code, it feels like this should maybe be in separate classes, rather than one Offsite class dealing with multiple different 'offsite' types

Gateway/Request/APMBuilder.php
```diff
@@ -82,6 +88,12 @@ public function build(array $buildSubject)
                     self::SOURCE_INSTALLMENT_TERMS => $method->getAdditionalInformation(InstallmentDataAssignObserver::TERMS),
                 ];
                 break;
+            case Truemoney::CODE:
```
jonrandy: Again, this implementation feels like it should be separate classes for each APM

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
```diff
@@ -0,0 +1,126 @@
...
+                return window.checkoutConfig.quoteData.store_currency_code;
+            },
+
+            /**
+             * Is method available to display
+             *
+             * @return {boolean}
+             */
+            isActive: function () {
+                return this.getOrderCurrency().toLowerCase() === 'thb' && this.getStoreCurrency().toLowerCase() === 'thb';;
```
jonrandy: Double semi-colon

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
```diff
@@ -0,0 +1,126 @@
+define(
```
jonrandy: Noticed that a lot of these `method-renderer` JS files are mostly the same. There'd probably be scope to make this more efficient? Feels very wrong to be repeating so much code

Controller/Callback/Offsite.php
jacstn: could you please explain more why? what is the reason for that? 99% of the code would be the same.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jacstn: i don't see solution for that actually. magento 2 is designed this way (it is used knockout js to push rendering from backend side to frontend side).
we should keep separated renderers because some of the parts are different. if you would see that all of them are objects it will be a mess to add in those renderers some extra conditions depends on what payment methods they concern.

guzzilar:
Haven't reviewed the code but would be nice if the pull request provides more information and screenshots. Something like 👇 
- How does the True Money payment page look like.
- The test in various cases, a screenshot showing admin order detail page with True Money payment method and what are expectations from each tests.
- A screenshot of admin payment setting page with a new payment method.
- Objective just looks the same as "Description of change", maybe Objective could be something that explain the reason why we should accept and merge this PR to the code base (aka. to convince readers)
- Maybe can also leave a reference link to Omise document (True Money page) as well.

Controller/Callback/Offsite.php
jonrandy: No, it wouldn't. You would have one parent class where all the stuff lives that is the same for every offsite method. Then you would have subclass for each particular offsite method where the unique could for that particualr offsite method would live. This keeps everything nicely separate, allows for easy changes (should any offsite method require something unusual), prevents the whole thing become monolithic  (full of ifs and switchblocks), and would allow for clean, easy addition/removal of new/outdated offsite methods

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jonrandy: Well, the code seems to just be calling a `define` function, and passing an array of strings and a function. The things that are passed seem almost the same in each `method-renderer` JS file. Couldn't we pull that array out to somewhere else, along with a function that generates the required function for passing in? So, in the `method-renderer` JS files, we would have something like:

```
define(DEPENDENCY_ARRAY, generateMethodRendererFunction('truemoney'))
```

My quick example is based upon looking at the Alipay and TrueMoney method renderer JS files - which are almost identical

Controller/Callback/Offsite.php
guzzilar: Actually seeing this and I feel the same. This class is getting fat and packed.
Quick one that make this feels wrong is by updating this line
```php
if ($order->getState() !== Order::STATE_PENDING_PAYMENT) { }
```
to
```php
if (! in_array($order->getState(), [ Order::STATE_PENDING_PAYMENT, Order::STATE_PROCESSING ])) { }
```
Just updated one line but we have to worry if this will affect the rest methods. While separating would be much more simpler (and bonus point, cleaner).

_I might need to consider this one for WC as well thou_

Gateway/Request/APMBuilder.php
```diff
@@ -82,6 +88,12 @@ public function build(array $buildSubject)
...
+                $paymentInfo[self::SOURCE] = [
+                    self::SOURCE_TYPE      => 'truemoney',
+                    self::SOURCE_PHONE_NUM => $method->getOrder()->getBillingAddress()->getTelephone(),
```
guzzilar: I'm not sure if using phone number from billing address would be practical for this True Money case.
Should we have an input to let user fill their phone number by themselves? Or at least having a choice to use a phone number from the billing address or enter a new one.

etc/adminhtml/system.xml
```diff
@@ -125,6 +125,21 @@
                         <config_path>payment/omise_offsite_installment/title</config_path>
                     </field>
                 </group>
+                <group id="omise_offsite_truemoney" translate="label" sortOrder="220" showInDefault="1" showInWebsite="1" showInStore="1">
```
guzzilar: sortOrder numbers are duplicated with Installment

Controller/Callback/Offsite.php
jacstn: @guzzilar according to our long `cancellation problem` talks we should get here only `pending` state on live server. `true money` introduced also `processing`. What other edge cases do you have in mind?

guzzilar: That is the problem, I don't know what it could possibly be as the one condition that used to applied for 3 payment methods was one thing, and now we altered that condition to make it support the 4th payment method.
Just raising a concern: "Will it affect or will it be any consequence for other payment methods as well?"

And that kind of concern normally happens when the code gets too fat (on the other hand, if we started to worry about this, then maybe our class is fat already).

guzzilar: _Btw, above are just to back up the idea of why I also feel that this should maybe be in separate classes. Not saying I have any particular edge case in mind._

Gateway/Request/APMBuilder.php
jacstn: @jonrandy could you present more info about why you think it should be done in separated files each one?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jacstn: that is reinventing how Magento is recommending to do it @jonrandy

jacstn: by any chance, did you find any example that supports your way of doing that changes on internet? magento official developer website?

Gateway/Request/APMBuilder.php
jonrandy: The comment directly relates to my previous explanation and comments by @guzzilar 

The reasoning is the same

Gateway/Request/APMBuilder.php
jacstn: that might be a good idea, I will consider to add it in the future PR.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jonrandy: It's common programming sense. I'm not changing anything about how Magento does things - simply removing unnecessary code duplication. We have very large blocks of code here that are all but identical except for a few strings. We would be calling the same Magento functions using the same arguments. Nothing changes except for more making it more maintainable, succinct code

Gateway/Request/APMBuilder.php
guzzilar: 👍 okie

Gateway/Request/APMBuilder.php
jacstn: Ok, I don't see any real problems it would solve now, yes.. it would nice to have, but definitely not in this PR, this potential refactoring concerns all payment methods.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jacstn: ok, thank you for your comment @jonrandy. any refactoring that concerns any other payment method will be done in separate PR.

Controller/Callback/Offsite.php
jacstn: ok, thanks @guzzilar @jonrandy .. refactoring that concerns other payment methods should be done in separate PR.

guzzilar: 👍 agreed

Controller/Callback/Offsite.php
jonrandy: ☝️

Controller/Callback/Offsite.php
jonrandy: I think this is bigger than just refactoring. It feels like sticking plaster code to make things work. It would be unclear to a new developer coming to this what is going on

Controller/Callback/Offsite.php
jacstn: @jonrandy why it feels wrong? `Feels wrong processing these two events via this code`
I don't understand where are your worries are coming from.
could you please describe real threats this code brings?
I mean.. what could happen.. and do you have any proofs (ie. omise.co documentation) that that worries could be confirmed.

Controller/Callback/Offsite.php
jacstn: @jonrandy @guzzilar this is how it works in magento 2 project I am responsible for:
I come in.. I add a solution to an existing architecture.
If architecture needs to be changed there are 2 scenarios:
1) if it architecture change is urgent then new PR is opened to change that architecture. And then this old PR needs to apply new architecture before is merged.
2) new pr is done, merged, architecture is changed later.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jacstn: BTW @jonrandy your bit emotional response and a quite big comment about the _common programming sense_ was not answering my question.
You answered:
_`It's common programming sense. I'm not changing anything about how Magento does things - simply removing unnecessary code duplication. We have very large blocks of code here that are all but identical except for a few strings - that's a really bad code smell. We would be calling the same Magento functions using the same arguments. Nothing changes except for more making it more maintainable, succinct code`_

I asked if you found anything how to do this stuff in Magento with `knockout JS`
in existing Magento architecture. it is easy to throw things.. and ask people .. `yeah, you check it!` I will just give you some ideas like:
`define(DEPENDENCY_ARRAY, generateMethodRendererFunction('truemoney')).`
because of _code smells_ for you.

jonrandy: Your question about Magento/Knockout examples did not make sense in relation to my comments, as my comments did not relate to Magento or Knockout. I am merely pointing out that in this part of the code we have very large sections of code that are being repeated. This is a situation that programming as a discipline in general seeks to remove.

I am not for one second advocating changing the way Knockout or Magento are being used, and I am not reinventing anything. I am merely suggesting that we improve our code to remove this unnecessary duplication. I will put together a related example to explain further

jonrandy: Upon closer inspection of the JS files there, it appears they are using Require JS. It would be relatively simple to split the 'method renderers' that we have into a class hierarchy with parent classes implementing the common code, and moving specifics into child classes. My proposed solution would work, but it circumvents the RequireJS mechanisms already available to us. Doing this spiltting would greatly improve the maintainability of the code, and reduce amount of code and effort required to add new payment methods on Magento 2. This is a worthwhile exercise and we should definitely do it.

Reference: https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/javascript/js_overview.html

jacstn: thank you @jonrandy I will have a look at it.

Controller/Callback/Offsite.php
guzzilar: @jacstn Did you test access this callback url with an order that paid with Alipay, Internet Banking, and Installment payment? Wonder what is going to happen.

As this original ` if ($order->getState() !== Order::STATE_PENDING_PAYMENT) { ... }` only allow an order with `pending` status to go through.

guzzilar: Actually, @jonrandy made a good point.
@jacstn is there any chance that this would conflict with Webhook feature, `charge.complete` event?

jacstn: I tested it.. everything except `Truemone` brings here `pending`, only `truemoney` firstly brings here first `pending` and then second callback `processing`.

guzzilar: @jacstn no, I mean, what if we "access" the url directly after being redirected here.
For example, I made a charge with Internet Banking, then the system brings me to the return uri: `https://www.mystore/callback/offsite/omise`. And the process went through as normal.

But somehow, I know this url and I attempt to access this one more time, what would happen in this case, with the order that has been paid with Internet Banking?

jacstn: don't see any problem it could bring to `complete` functionality.

jonrandy:
What ignoring the second hit on the URL and just dealing with just the
webhook as normal? Also, have we asked the back end devs why this second
hit on the return URL is happening?

jonrandy:
*what about

jacstn:
@jonrandy please.. again it would be awesome and you would not reply by email, just reply to comment, instead of posting random replies.. as email do

Controller/Callback/Offsite.php
jacstn: webhook is different `controller/action`

jacstn: so do you mean @guzzilar:
1. payment is made
2. status is updated with a callback as `pending`
3. webhook come, and payment is completed
4. again callback comes from point 2 with the same, `pending` or `processing`

if this will happen, the merchant will have the wrong status of payment according to my knowledge, the only good status will be visible in omise.co

jonrandy: What about if we just ignoried the second hit on the URL and just deal with just the webhook as normal? Also, have we asked the back end devs why this second hit on the return URL is happening? Is there a reason for it?

guzzilar:
Screen Shot 2562-06-19 at 17 50 08
Just one quick thing, there is no "Test mode" label at True Money payment method.

guzzilar:
Screen Shot 2562-06-19 at 18 02 36
One more quick note, got double notes at the order that is paid with True Money.

jonrandy:
> Screen Shot 2562-06-19 at 17 50 08
> Just one quick thing, there is no "Test mode" label at True Money payment method.

Good catch. We should probably move this warning out of the individual payment methods and just have one on the page - that is less work overall, keeps it nicely separate, and prevents it being forgotten on new methods when they are added

jonrandy:
> Screen Shot 2562-06-19 at 18 02 36
> One more quick note, got double notes at the order that is paid with True Money.

Possibly related to the double hit on the `return_uri`?

view/frontend/web/template/payment/offsite-truemoney-form.html
```diff
@@ -0,0 +1,61 @@
...
+               },
+               value: getCode(),
+               checked: isChecked,
+               click: selectPaymentMethod,
+               visible: isRadioButtonVisible(),
+               enable: isActive()" />
+        <label data-bind="attr: {'for': getCode()}" class="label">
+            <span data-bind="text: getTitle()"></span>
+        </label>
+        <div data-bind="visible: isSandboxOn()" class="page messages">
```
jonrandy: This is definitely in the wrong place, we shouldn't be adding another warning for every new payment method. Test/Sandbox mode always applies to all Omise payment methods. The warning should really be a single one elsewhere on the page. If we really do want to put a warning on every payment method though, we should at least be calling the same "isSandbox" method to decide whether or not to display it. All this code duplication is bad.

I guess this relates back to the other refactoring we discussed before.

jacstn: what if a merchant has PayPal on or cash on delivery. We should show exactly which payments are in test mode.

jonrandy: The message is only relevant for the merchant. Telling them that 'Omise payment methods' are in test mode is definitely enough info. We should credit the merchant with some intelligence

jonrandy: It's about reminding them that test mode is on, so they don't leave it on when going live

jacstn: then if it is the only a reminder, that is fine with the current implementation, 

and also it covers the situation when a merchant will really forget. The current implementation is way better than showing a single message.

jonrandy: The current implementation is overly complex, adds a maintenance burden, and treats the merchant like an idiot by telling them again and again and again

jacstn: your aggressiveness in public or private debate is something you should consider.
if you feel hard to review PRs in a civilised way I will stop sending you requests.

jonrandy: No aggression was intended. It was merely a statement of fact. No capital letters, no bashing of keyboard 🌻

guzzilar: I agreed on showing one message somewhere is better than showing in every payment methods. Just for 2 main reasons.

1. Complexity of the code, as this case may be a good example that we can easily forget about adding this label, also the code has been implemented everywhere, it's hard to maintenance and risk of human error. 

2. As user experience, having "test mode" in each payment methods may raise a question:
"Is it possible to turn `Alipay` to test mode while keep the rest in live mode." As the design is indicating that "test mode" is under and controlled by each individual payment method.

If we talking about achieving on raising an awareness of "accidentally leaving the test mode on", I think both are achieving the same goal, but having one place to maintain would give more benefit than another.

guzzilar: However, do we have any conclusion on this? : )))
What action should we take? Want to fix this one before this PR? or we may submit another PR to improve this code, maybe?

jonrandy: Another thing that occurred to me over the weekend was that putting the warnings actually next to the payment methods like this will mean that they only see the warnings once they have gone through all the previous checkout steps. The likelihood of them leaving test mode on in this case is for that reason probably higher than if the warning is outside of this area and always visible

guzzilar: 🤚 One thing that not related to the above but I just realised. 
If nothing has changed, I think we should stick with `isTest()` (or `isTestMode()`) rather than `isSandboxOn`. As normally Omise uses `Test` term widely and almost everywhere.

jacstn: Thanks for all comments, but for now, it stays at it is.

guzzilar: @jonrandy good point. 👍

jonrandy: > I agreed on showing one message somewhere is better than showing in every payment methods. Just for 2 main reasons.
> 
>     1. Complexity of the code, as this case may be a good example that we can easily forget about adding this label, also the code has been implemented everywhere, it's hard to maintenance and risk of human error.
> 
>     2. As user experience, having "test mode" in each payment methods may raise a question:
>        "Is it possible to turn `Alipay` to test mode while keep the rest in live mode." As the design is indicating that "test mode" is under and controlled by each individual payment method.
> 
> 
> If we talking about achieving on raising an awareness of "accidentally leaving the test mode on", I think both are achieving the same goal, but having one place to maintain would give more benefit than another.

Totally agree on 1) - the human error thing already almost happened as the warning was initially forgotten for True Money. As for 2) I agree also - having a warning for every method does imply that it may be possible to put each method in test mode individually

jacstn: @guzzilar yes.. your function name might be better.
but it is consistent with what we have in the admin panel.
we use `Sandbox`
I used to display here `Sandbox` too, but it was changed by support team. this message.
that is why is sandbox: 
image
 We could consider changing all of that to testmode including panel info.

jacstn: for somebody who work with the code it might be easier to recognize what exactly is he dealing with when we leave `sandbox`

text messages are always fully customizable,
in polish, we would use `sandbox` for instance if you would translate it.
not test mode.

guzzilar: Just to make entire Omise products consistent (test key, test account, test mode, as we mentioned in documents and Dashboard) I still vote that we should consider replace all `Sandbox` to `Test mode` but in the future PR. ;)

jacstn: ok @guzzilar this cycle will do refactoring.

Controller/Callback/Offsite.php
jacstn: indeed we do ignore, but still, we need to redirect.

Controller/Callback/Offsite.php
```diff
@@ -60,6 +61,10 @@ public function execute()
     {
         $order = $this->session->getLastRealOrder();
 
+        if ($order->getState() === Order::STATE_PROCESSING) {
```
guzzilar: @jacstn I'm still not sure if this is a good idea to redirect `Order::STATE_PROCESSING` order to the `PATH_SUCCESS` whenever user enters this callback url.

Just wonder for only one question,
- Will any action from Magento be triggered when user enters the page? (i.e. is there any script that would be executed when the page is opened?)

If nothing at all, Magento just simply renders a HTML success page, then maybe ok... (still not sure, but well, no question further).

guzzilar: Also, this condition should be checked after the below condition.
```php
if (! $order->getId()) {
    $this->messageManager->addErrorMessage(__('The order session no longer exists, please make an order again or contact our support if you have any questions.'));
	
    return $this->redirect(self::PATH_CART);
}
```

jacstn: yes.. @guzzilar this callback when successfully done change state of order to `processing` than when already `processing`, no need to do anything.

this action only redirects to success page.

jonrandy: Any news on the fix to prevent this 'double redirect' even happening? Feels like we are addressing a symptom rather than the cause

jacstn: backend team did not find any causes yet @jonrandy .

PR #204, opened 2019-06-16
Proposal / idea – refactoring the offsite callback class.

guzzilar:
**⚠️ Disclaimer:**
**This is not a working code and never been tested before. As well as, not-so-polished method and variable names. Simply just an idea for a discussion.**

According to the conversation at the pull request #201, 
https://github.com/omise/omise-magento/pull/201/files#r292773456

Here is an idea & a proposal on refactoring the offsite callback class.
The objective of this refactoring is to
- To decouple a big-fat code in one `execute` method
- To make it unit-test friendly. As if we are able to decouple code into those small methods, we would as well, write a unit test more easier.
- Improve its readability.

### DIFF
--- a/Controller/Callback/Base.php
+++ b/Controller/Callback/Base.php
@@ -1,0 +1,261 @@
+<?php
+namespace Omise\Payment\Controller\Callback;
+
+use Magento\Sales\Model\Order;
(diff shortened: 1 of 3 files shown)


jonrandy (commented):
:)

aashishgurung:
Stale PR.

PR #205, opened 2019-06-21
Change to display an error message using ErrorMessageMapper

tamada1:
#### 1. Objective

Currently, credit card payment error messages are not displayed correctly.
It should be changed to display an error message according to the error code.

#### 2. Description of change

Change to use ErrorMessageMapper to display an error message.
As of the present, if you do not use ErrorMessageMapper, the function of \Magento\Payment\Gateway\Command\GatewayCommand::processErrors will display only the following messages.
'Transaction has been declined. Please try again later.'
Also, it is not desirable to manipulate the contents of the error message before passing it to the ErrorMessageMapper.
We should pass the error code as is.

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 2.2.8.
- **PHP version**:  7.1.30.

**✏️ Details:**

In the One-Page Checkout payment step, enter and confirm incorrect credit card information.

#### 4. Impact of the change

The error message for credit card payments is changed.

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/etc/omise_error_mapping.xml
+++ b/etc/omise_error_mapping.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+<mapping xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Payment:etc/error_mapping.xsd">
+    <message_list>
+        <message code="failed_fraud_check" translate="true">Card was marked as fraudulent.</message>
+        <message code="failed_processing" translate="true">General payment processing failure.</message>
+        <message code="insufficient_balance" translate="true">Insufficient funds in the account or the card has reached the credit limit.</message>
+        <message code="insufficient_fund" translate="true">Insufficient funds in the account or the card has reached the credit limit.</message>
+        <message code="invalid_account_number" translate="true">Valid account for payment method not found.</message>
+        <message code="invalid_account" translate="true">Valid account for payment method not found.</message>
+        <message code="invalid_security_code" translate="true">Security code invalid or card did not pass preauthorization.</message>
+        <message code="payment_cancelled" translate="true">Payment cancelled by payer.</message>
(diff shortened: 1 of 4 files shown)


jacstn:
@tamada1 thank you for this PR, we will have look at it.

etc/omise_error_mapping.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<mapping xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Payment:etc/error_mapping.xsd">
+    <message_list>
```
jacstn: Do we need from any UI perspective to display so detailed info to user?
like `Card was marked as fraudulent`

tamada1: I agree that we do not need to display a very detailed error message to the user.
However, I do not think that it is correct that all errors return a single message as in the current implementation.
What do you think of using i18n to replace messages for users?

jonrandy: I would think if if a user's card had been marked as fraudulent, they would want to know about it! I certainly would

jonrandy: If error messages being shown to the user are not currently translated to the store's current language, I think they should be

jacstn:
tested it.. works well. thanks @tamada1 

We will wait for other guys PR approval.

etc/omise_error_mapping.xml
jacstn: @jonrandy I will add the translation in the next PR.

jacstn:
@guzzilar @jonrandy, do you want to add here some comment?

etc/omise_error_mapping.xml
jonrandy: It would make more sense to add it to this PR really. It is directly related

PR #207, opened 2019-07-09
Edit phone number on checkout-page when making True Money payment

jacstn:
#### 1. Objective

Current True Money Payment implementation takes phone number required for making a payment from a customer's profile.
If a user wanted to use a different number than saved in the profile he had to change it directly in profile settings before making payment.
This PR adds an option to modify phone number used for True Money Payment directly on the checkout form.

#### 2. Description of change

Added input form element and related JS functionality to allow to change a phone number on UI side.
Added Magento 2 data observers to capture new phone number and send it to payment gateway.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

Make payment using True Money option in both cases:
1. Edit phone number before making payment
2. Leave phone number unchanged.

#### 4. Impact of the change

User can modify phone number that is used to make True Money Payment using the following input.

image

If input remains blank, then phone number from the saved profile is used.
Phone number saved in the user's profile is displayed as `placeholder`.

#### 5. Priority of change

Normal.

#### 6. Additional Notes

N/A


Observer/TruemoneyDataAssignObserver.php
```diff
@@ -0,0 +1,46 @@
+<?php
+namespace Omise\Payment\Observer;
+
+use Magento\Framework\Event\Observer;
+use Magento\Payment\Observer\AbstractDataAssignObserver;
+use Magento\Quote\Api\Data\PaymentInterface;
+
+class TruemoneyDataAssignObserver extends AbstractDataAssignObserver
```
jonrandy: Again, as before we have a lot of repeated code in the `DataAssignObserver`s for a lot of these methods. Three of them (offsite ones?) all feature identical `execute` methods. We should refactor to avoid the repeated code - perhaps inherit from a `OffsiteDataAssignObserver` class?

Observer/TruemoneyDataAssignObserver.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * @var string
+     */
+    const PHONE_NUM   = 'truemoney_phone_num';
```
jonrandy: Naming feels wrong since this isn't actually a phone number

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
```diff
@@ -5,18 +5,16 @@ define(
         'mage/storage',
```
jonrandy: We're building here on top of code that is due for quite large refactoring. Wouldn't it be less work overall to do the refactoring first? This way just creates more work. Kind of academic really since this is already done though, I guess

Observer/TruemoneyDataAssignObserver.php
jacstn: how would you name it?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-truemoney-method.js
jacstn: Will be subject of future refactoring in some cool-off period.
Business cases are my biggest priority now.
Also, I don't see a threat of any unnecessary complexity in the future.

Observer/TruemoneyDataAssignObserver.php
jacstn: Thanks, This pr implements functionality in current architecture.
Your suggestions will be the subject to future tasks.

Observer/TruemoneyDataAssignObserver.php
guzzilar: Maybe
```php
const PHONE_NUMBER = 'truemoney_phone_number';
```

Because `phone_number` has been used all over the code and `phone_num` is just only in this place.

jonrandy: Maybe something like PHONE_NUMBER_FIELD? The name phone_num or phone_number suggests the constant should be a phone number, which it isn't

jacstn: changed to `num` => `number`

jacstn: no reason to add there `FIELD` suffix.
the current name is understandable and keeps a user's phone number.

jonrandy: It doesn't hold the user's phone number. The constant holds the name of the field that holds the user's phone number

jacstn: it is hard not to agree.. .but we keep this naming convention in the entire plugin.

guzzilar: My bad, it was started by me adding a constant to reflect with its field name just to make it looks obvious to see. But didn't think of it, that is not actually contain a value, just a field name. 
i.e.
```php
    self::SOURCE_TYPE => 'alipay',
];
```

Think of this again, calling `Omise\Payment\Gateway\Request\AlipayBuilder::SOURCE_TYPE` and get `source_type`, look a bit off.
Please feel free to fix the entire convention if necessary.

jacstn: will definitely consider in the future.

jonrandy: > it is hard not to agree.. .but we keep this naming convention in the entire plugin.

We should always endeavour to fix anything with code as we find the problems. Putting stuff off like this is a bad way to go about things. If we keep 'considering things' for the future, they will almost certainly not get done. I've seen this many, many times in the past. It gets to a point where either all the proposed changes are forgotten about, or the 'can kicking' has gotten to the point where the amount of 'technical debt' incurred is too intimidating and no-one wants to do the fixing any more as it is a big job.

Less work is required overall if we fix things as we find them. We should, as far as possible, keep our house tidy

jacstn: the reason is that I don't see it as any big thing. That is the reason.
Business cases are way more important for me then naming convention that is not maybe the best but we still kept it for many PRs unnoticed, and for me, it is very completely understandable.

jacstn: All of the subjects to refactor are for me future cases, don't see any of them as a threat for any complexity. yes, we have more javascript files, but a number of javascript files are a way more minor thing than allowing our customers to make money on ie. Convenience Store Payment.

That all refactorings in Magento 2 will be subjects of future stuff after delivering all business cases.

jonrandy: This is precisely my point, and in no way am I just referring to this particular issue with naming convention. We don't do things as we don't consider them 'big things' - and before we know it, we have a huge pile of 'little' things that all really need attention. I have also pointed out in this PR the code has been added on top of other code that is in rather large need of refactoring, and this refactoring too was 'considered' for the future after previous PRs, but has not been done. This has the effect of increasing the amount of technical debt, and increasing the total amount of work that has to be done on the project overall

jonrandy: By not doing the refactoring as we go, we are creating more work - and potentially delaying/complicating future business cases.

It isn't something we should do later as a chore, it should be an ongoing part of the development process

jacstn: well, that's are your feelings.
I don't see it as any threat of any complexity, and we have different opinions in naming, for instance, something you call: a `huge` pile of 'little' things, it is for me it is a `huge` exaggeration.

is there anything that holds you guys from approving this PR?

jonrandy: I was talking in the general case about where this style of development usually leads. I've seen this happen too many times before to ignore it.

Presentations from Omise's CTO also encourage us to work in this manner

jacstn: ok, is there anything that holds you guys from approving this PR?

Gateway/Request/APMBuilder.php
```diff
@@ -91,7 +92,7 @@ public function build(array $buildSubject)
             case Truemoney::CODE:
                 $paymentInfo[self::SOURCE] = [
                     self::SOURCE_TYPE      => 'truemoney',
-                    self::SOURCE_PHONE_NUM => $method->getOrder()->getBillingAddress()->getTelephone(),
+                    self::SOURCE_PHONE_NUM => $method->getAdditionalInformation(TruemoneyDataAssignObserver::PHONE_NUMBER),
```
guzzilar: We don't want to replace `SOURCE_PHONE_NUM` to `SOURCE_PHONE_NUMBER` as well?

view/frontend/web/css/styles.css
```diff
@@ -94,4 +94,9 @@
...
 #omise_offsite_internetbankingForm img.kbank, #omise_offsite_installmentForm img.kbank {
     background: url('../images/kbank.svg')rgb(247, 226, 226);
-}
\ No newline at end of file
+}
+
+#omise_offsite_truemoneyphoneNumber {
+    width: 225px;
+    margin-bottom:18px;
+}
```
guzzilar: These numbers may affect with the mobile or tablet screen, would be nice if we also test on those 2 screens if it fits properly.

Gateway/Request/APMBuilder.php
jacstn: i missed that.. :|

view/frontend/web/css/styles.css
jacstn: that `width` style is copied from credit card, margin-bottom looks normal on the phone.

PR #212, opened 2019-07-26
Page configuration - fix Schema Namespace in checkout

jacstn:
#### 1. Objective

This PR fix Checkout Configuration page.

#### 3. Quality assurance

- **PHP version**: 7.0.16.

**✏️ Details:**

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/view/frontend/layout/checkout_index_index.xml
+++ b/view/frontend/layout/checkout_index_index.xml
@@ -40,9 +40,6 @@
                                                                         <item name="omise_offsite_truemoney" xsi:type="array">
                                                                             <item name="isBillingAddressRequired" xsi:type="boolean">true</item>
                                                                         </item>
-                                                                        <item name="omise_offsite_conviniencestore" xsi:type="array">
-                                                                            <item name="isBillingAddressRequired" xsi:type="boolean">true</item>
-                                                                        </item>
                                                                     </item>
                                                                 </item>
                                                             </item>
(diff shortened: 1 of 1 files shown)


guzzilar:
@jacstn can you leave some detail on what is the error and how does this PR solve that error?, any reference link would be ok too.

jacstn:
Sure @guzzilar, once it's ready for review, it will have all info.

jacstn:
thanks for review @jonrandy although the fix is working well, I have not found out yet if there are any potential problems it might cause in current installations once upgraded.

So I will not integrate it soon.
No priority with this one.

jacstn:
Customer reported problem:
```
We found a bug in the Omise_Payment extension's layout file. This was breaking the layout of the checkout page by using the wrong property in the page node.

Currently, it is having the following code in its vendor/omise/omise-magento/view/frontend/layout/checkout_index_index.xml file.

<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" layout="1column" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/layout_generic.xsd"> 

We checked and compared it with other files and found that it should be
<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" layout="checkout"
      xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">

Please check the xsi:noNamespaceSchemaLocation property. It should be urn:magento:framework:View/Layout/etc/page_configuration.xsd where as the extension is using urn:magento:framework:View/Layout/etc/layout_generic.xsd 
The current version of the module is "2.7.0"

Please let us know about this.
```

jacstn:
For instance, what will happen for Magento 2 setups with built themes, and customized around current layout configuration.

guzzilar:
@jacstn ah sorry, I didn't see that it's still in progress. But thanks for the explanation 👍

aashishgurung:
Stale PR.

PR #215, opened 2019-08-30
Error when change Omise Account - Fix

jacstn:
This PR adds fix.

An error occurred when a user changed Omise account.

#### 1. Objective

An error occurred because Magento was using on checkout customer ID that was created using old account's data (Private Key).
When Merchant's Private Key has been changed and used from different Omise Account, then customer id did not match new merchant's data.
Customer Object was an instance of the error. All operations on this object were throwing exceptions.

#### 2. Description of change

This PR adds a mechanism that checks if an instance of the `customer API object` is Error, if yes, then it refreshes omise customer data. This PR assumes that 

*Note:*
In case of changing omise account, all saved customer's credit card information becomes unavailable.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

1. Create a Magento customer's account/profile or use an existing one.
2. Make payment (make sure you DO NOT check the box to save credit card)
3. Payment should be successful.

4. Make payment (make sure you check the box to save credit card)
5. Payment should be successful, the card's data should be stored.

6. Make payment with saved credit card
7. Payment should be successful, card should be still stored.

**Change Omise Keys to different Omise Account.**

8. Login to Magento account used in Point 1.
9. Make payment (make sure you DO NOT check the box to save credit card)
10. Payment should be successful.

11. Make payment (make sure you check the box to save credit card)
12. Payment should be successful, the card's data should be stored.

13. Make payment with saved credit card
14. Payment should be successful, card should be still stored.

#### 4. Impact of the change

This PR change Customer ID generated using different omise account.
Stored credit cards from all customers (saved using previous omise account) will be unavailable in case of changing omise account.
These cards are not deleted, they will be available again if previous omise account keys are again used.

#### 5. Priority of change

Immediate.

#### 6. Additional Notes

N/A

### DIFF
--- a/Model/Customer.php
+++ b/Model/Customer.php
@@ -172,12 +172,15 @@ class Customer
      */
     protected function initializeObject()
     {
-        if ($this->getId()) {
-            $customerAPI = $this->customerAPI->find($this->getId());
+        $customerId = $this->getId();
+
+        if ($customerId) {
+            $customerAPI = $this->customerAPI->find($customerId);
 
             if ($customerAPI instanceof \Omise\Payment\Model\Api\Error) {
                 // instance of Error if merchants Omise account and key has been changed
                 $this->create();
+                //after re-creating customer, it's ID has beenchanged, so we need to find API using new customer ID, so get it again.
                 $customerAPI = $this->customerAPI->find($this->getId());
             }
 
--- a/Model/Customer.php
+++ b/Model/Customer.php
@@ -97,9 +97,12 @@ class Customer
      *
      * @return self
      */
-    public function create(array $params)
+    public function create()
     {
-        $this->customerAPI = $this->customerAPI->create($params);
+        $this->customerAPI = $this->customerAPI->create([
+            'email'       => $this->magentoCustomer->getEmail(),
+            'description' => trim($this->magentoCustomer->getFirstname() . ' ' . $this->magentoCustomer->getLastname())
+        ]);
 
         $this->magentoCustomer->setData(self::OMISE_CUSTOMER_ID_FIELD, $this->customerAPI->id);
         $this->magentoCustomerResource->saveAttribute($this->magentoCustomer, self::OMISE_CUSTOMER_ID_FIELD);
@@ -117,10 +120,7 @@ class Customer
     public function addCard($cardToken)
     {
         if (! $this->getId()) {
-            $this->create([
-                'email'       => $this->magentoCustomer->getEmail(),
-                'description' => trim($this->magentoCustomer->getFirstname() . ' ' . $this->magentoCustomer->getLastname())
-            ]);
+            $this->create();
         }
 
         $this->customerAPI = $this->customerAPI->update(['card' => $cardToken]);
@@ -163,7 +163,6 @@ class Customer
     public function getLatestCard()
     {
         $cards = $this->cards(['order' => 'reverse_chronological']);
-
         return $cards['total'] > 0 ? $cards['data'][0] : null;
     }
 
@@ -174,7 +173,15 @@ class Customer
     protected function initializeObject()
     {
         if ($this->getId()) {
-            $this->customerAPI = $this->customerAPI->find($this->getId());
+            $customerAPI = $this->customerAPI->find($this->getId());
+
+            if ($customerAPI instanceof \Omise\Payment\Model\Api\Error) {
+                // instance of Error if merchants Omise account and key has been changed
+                $this->create();
+                $customerAPI = $this->customerAPI->find($this->getId());
+            }
+
+            $this->customerAPI = $customerAPI;
         }
     }
 }


Model/Customer.php
```diff
@@ -174,7 +173,15 @@ public function getLatestCard()
     protected function initializeObject()
     {
         if ($this->getId()) {
-            $this->customerAPI = $this->customerAPI->find($this->getId());
+            $customerAPI = $this->customerAPI->find($this->getId());
+
+            if ($customerAPI instanceof \Omise\Payment\Model\Api\Error) {
+                // instance of Error if merchants Omise account and key has been changed
```
jacstn: this comment is understandable?

Model/Customer.php
```diff
@@ -117,10 +120,7 @@ public function create(array $params)
     public function addCard($cardToken)
     {
         if (! $this->getId()) {
-            $this->create([
-                'email'       => $this->magentoCustomer->getEmail(),
-                'description' => trim($this->magentoCustomer->getFirstname() . ' ' . $this->magentoCustomer->getLastname())
-            ]);
```
jacstn: Easier if these params are inside the function rather then being passed in function parameters.
Moved it to the function.

Model/Customer.php
jacstn: added a comment because I have no idea how I would name variables/function to make it (code) describing what it is and why it is here.

jonrandy:
>2. Description of change
>
>This PR adds a mechanism that checks if an instance of the customer API object is Error, if yes, then >it refreshes omise customer data. This PR assumes that
>
>Note:
>In case of changing omise account, all saved customer's credit card information becomes >unavailable.

What does the PR assume? You appear to have ended mid sentence?

Model/Customer.php
```diff
@@ -174,7 +173,15 @@ public function getLatestCard()
     protected function initializeObject()
     {
         if ($this->getId()) {
-            $this->customerAPI = $this->customerAPI->find($this->getId());
+            $customerAPI = $this->customerAPI->find($this->getId());
```
jonrandy: Maybe we could just call `getId()` once and use the result

jacstn: > > 1. Description of change
> > 
> > This PR adds a mechanism that checks if an instance of the customer API object is Error, if yes, then >it refreshes omise customer data. This PR assumes that
> > Note:
> > In case of changing omise account, all saved customer's credit card information becomes >unavailable.
> 
> What does the PR assume? You appear to have ended mid sentence?

@jonrandy so what I meant was:

When you change omise account, register new account with omise, generate private key and public key, then if you change that key in magento shop in payment options, then when customer comes back to the shop he cannot see all his credit cards that were saved before.

jacstn:
> > 1. Description of change
> > 
> > This PR adds a mechanism that checks if an instance of the customer API object is Error, if yes, then >it refreshes omise customer data. This PR assumes that
> > Note:
> > In case of changing omise account, all saved customer's credit card information becomes >unavailable.
> 
> What does the PR assume? You appear to have ended mid sentence?

@jonrandy so what I meant was:

When you change omise account, register new account with omise, generate private key and public key, then if you change that key in magento shop in payment options, then when customer comes back to the shop he cannot see all his credit cards that were saved before.

Model/Customer.php
jacstn: 👍

guzzilar:
While this pull request solves the issue that has been raised, I'm afraid that it may introduces a new issue where the code is secretly replacing user's Omise customer id when it fails to retrieve Omise Customer object without alerting merchants for this consequence – and there is no way we can restore its previous data back (paring between a particular Magento customer id with Omise Customer id).

This may be overthinking but, what do you guys think? Should we have any mechanism to say, "this particular Magento customer id is paring with 2 Omise Customer IDs and each id will be used based on a store's credential"? 
 
Alternative way that I can think of would be to alert merchant when they update their keys that their customers' saved-card will be affected by this change.

jacstn:
@guzzilar the business case scenario is only for people who will sell their shop, the new owner needs to create a new omise account.
Do you think text-info/warning in admin panel would be enough?

guzzilar:
@jacstn I think the above concern could be applied to the business case you mentioned as well.
Merchant [A] sells their business to merchant [B], then merchant [B] didn't aware of that all the customer data (in Omise system) will be replaced and all customers' saved cards would be replaced.

Also, I think there were some amount of queries in the past where merchant ask Omise if Omise can import customer data from one account to another (when there is a change on their business main account / owner) – would be worth to check?

jacstn:
ok @guzzilar .. what is the plan..? check.. then how it will affect this pr, what scenarios do you see here?

this PR is very edge situation, as you can imagine.
reported so far by 1 customer.. ever

Model/Customer.php
guzzilar: I think its interface is a bit strange where the method name says `"addCard"` then check if not `"this->getId()"` then `"this->create()"`.
```php
public function addCard($cardToken)
{
    if (! $this->getId()) {
        $this->create();
    }

   ...
}
```

jacstn: it is your code... @guzzilar

guzzilar:
@jacstn So I wonder if we should silently replace customer data without giving any alert or acknowledge merchant in the first place. Replacing data is fine, but should we somehow inform merchant first that replacing keys will have its consequence of losing customer data? – something like that.

jacstn:
@guzzilar this info will be added when a reorganisation of admin panel will be done. In separated PR

Model/Customer.php
guzzilar: I was also feeling off for this part too. But still, can't figure it out where is the proper place to have this one (unless we imitate OmisePHP library by having nested-classes – something like, `$this->customer->card->add()`)

Issue #189, opened 2019-05-06
[BUG] Prevent retrieve authentication while module was not configured or key not added yet


tuyennn:
### Steps to reproduce
- Magento 2.2.7
- Fresh install module
- Payment was not configured or added any key

### Expected result
- From checkout or every constructor class will not call retrieve or authentication data or anything else, unless the module enabled(key added)

### Actual result
- Got Exception on some __construct while the authentication was called but module was not used(key not added)

![image](https://user-images.githubusercontent.com/4010395/57218375-5989a200-701f-11e9-9ead-3670f80c3832.png)

![image](https://user-images.githubusercontent.com/4010395/57218397-6ad2ae80-701f-11e9-8733-922218610fdd.png)

jacstn:
ok, thank you for your bug report. 
Will investigate and fix it asap.

tuyennn:
@jacstn what is the status of this issue? I have to temporary disable your payment gateway on all our development, staging environments.

jacstn:
fixed.

PR #216, opened 2019-09-02
Handle exception when getting capabilities with invalid Omise Keys

jacstn:
#### 1. Objective

This fixes unhandled exception when omise keys are wrong or not set in the admin panel.

**Related information**:
Related issue(s): #189 

#### 2. Description of change

Added `try/catch` when getting capabilities.
Return particular capabilities only if the capabilities API object is not null. 

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

Remove omise keys, make sure application does not crash when going to the checkout page.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/Model/Api/Capabilities.php
+++ b/Model/Api/Capabilities.php
@@ -8,7 +8,7 @@ class Capabilities extends BaseObject
 {
     private $capabilities;
 
-    public function __construct(\PSR\Log\LoggerInterface $log)
+    public function __construct()
     {
         try {
             $this->capabilities = OmiseCapabilities::retrieve();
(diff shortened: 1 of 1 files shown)


Model/Api/Capabilities.php
```diff
@@ -8,9 +8,13 @@ class Capabilities extends BaseObject
 {
     private $capabilities;
 
-    public function __construct()
+    public function __construct(\PSR\Log\LoggerInterface $log)
```
jonrandy: Is `$log` used?

PR #222, opened 2019-09-09
Hide installment companies using Admin Panel

jacstn:
#### 1. Objective

This PR enables the possibility of hiding installment companies by the merchant in the admin panel.

#### 2. Description of change

Added new Magento config entries representing each of installment providing company.
Added in admin panel UI list of options to enable/disable Installment option.
Added in JS/HTML on checkout page functionality to hide installment options depends on admin panel settings.

#### 3. Quality assurance

- **PHP version**: 7.0.16.

**✏️ Details:**

To manually test this feature enable/disable installment options in the admin panel.
Check if disabled options disappear from the checkout page. 
Make a successful payment with enabled options.

#### 4. Impact of the change

This PR adds a new section under the Installment Payment Option on the Admin Panel.
The section is named _Advanced_ and is used by a merchant can disable unwanted installment providers.
By default, all companies are enabled.

image

If a merchant enables Installment Payment in Admin panel, but turn off all installment companies, _There are no installment options available_ text will be displayed on the checkout page. Also selecting Installment on the checkout page will be disabled.

image

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A

### DIFF
--- a/Model/Config/Installment.php
+++ b/Model/Config/Installment.php
@@ -11,8 +11,8 @@ class Installment extends Config
     const CODE = 'omise_offsite_installment';
 
     const KBANK       = 'omise_offsite_installment_kbank';
-    const KRUNGTHAI   = 'omise_offsite_installment_krungthai';
-    const FIRSTCHOICE = 'omise_offsite_installment_fc';
+    const KRUNGTHAI   = 'omise_offsite_installment_ktc';
+    const FIRSTCHOICE = 'omise_offsite_installment_first_choice';
     const BBL         = 'omise_offsite_installment_bbl';
-    const KRUNGSRI    = 'omise_offsite_installment_krungsri';
+    const KRUNGSRI    = 'omise_offsite_installment_bay';
 }
--- a/Model/Config/Installment.php
+++ b/Model/Config/Installment.php
@@ -9,4 +9,10 @@ class Installment extends Config
      * @var string
      */
     const CODE = 'omise_offsite_installment';
+
+    const KBANK       = 'omise_offsite_installment_kbank';
+    const KRUNGTHAI   = 'omise_offsite_installment_krungthai';
+    const FIRSTCHOICE = 'omise_offsite_installment_fc';
+    const BBL         = 'omise_offsite_installment_bbl';
+    const KRUNGSRI    = 'omise_offsite_installment_krungsri';
 }
--- a/Model/Config/Installment.php
+++ b/Model/Config/Installment.php
@@ -13,25 +13,25 @@ class Installment extends Config
     /**
      * @var string
      */
-    const KBANK = 'omise_offsite_installment_kbank';
+    const KBANK = CODE + '_kbank';
 
     /**
      * @var string
      */
-    const KRUNGTHAI = 'omise_offsite_installment_ktc';
+    const KRUNGTHAI = CODE + '_ktc';
 
     /**
      * @var string
      */
-    const FIRSTCHOICE = 'omise_offsite_installment_first_choice';
+    const FIRSTCHOICE = CODE + '_first_choice';
 
     /**
      * @var string
      */
-    const BBL = 'omise_offsite_installment_bbl';
+    const BBL = CODE + '_bbl';
 
     /**
      * @var string
      */
-    const KRUNGSRI = 'omise_offsite_installment_bay';
+    const KRUNGSRI = CODE + '_bay';
 }
--- a/Model/Config/Installment.php
+++ b/Model/Config/Installment.php
@@ -10,9 +10,28 @@ class Installment extends Config
      */
     const CODE = 'omise_offsite_installment';
 
-    const KBANK       = 'omise_offsite_installment_kbank';
-    const KRUNGTHAI   = 'omise_offsite_installment_ktc';
+    /**
+     * @var string
+     */
+    const KBANK = 'omise_offsite_installment_kbank';
+
+    /**
+     * @var string
+     */
+    const KRUNGTHAI = 'omise_offsite_installment_ktc';
+
+    /**
+     * @var string
+     */
     const FIRSTCHOICE = 'omise_offsite_installment_first_choice';
-    const BBL         = 'omise_offsite_installment_bbl';
-    const KRUNGSRI    = 'omise_offsite_installment_bay';
+
+    /**
+     * @var string
+     */
+    const BBL = 'omise_offsite_installment_bbl';
+
+    /**
+     * @var string
+     */
+    const KRUNGSRI = 'omise_offsite_installment_bay';
 }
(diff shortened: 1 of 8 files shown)


view/frontend/web/template/payment/offsite-installment-form.html
```diff
@@ -25,7 +25,14 @@
                 </div>
             </div>
         </div>
-        <div data-bind="visible: isActive() && orderValueTooLow()" class="page messages">
+        <div data-bind="visible: !isAnyInstallmentOptionActive()" class="page messages">
+            <div role="alert" class="messages">
+                <div class="message-warning warning message">
+                    <div data-bind="i18n: 'There are no any installment options available'"></div>
```
jacstn: is that correct english?
_There are no any installment options available_

Model/Config/Installment.php
```diff
@@ -9,4 +9,29 @@ class Installment extends Config
      * @var string
```
jonrandy: Docblock should be `@const`? Saw [this](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-javascript.html#frontend-dev-guide-javascript-js-coding-docblock-tagreference)

Model/Ui/InstallmentConfigProvider.php
```diff
@@ -0,0 +1,61 @@
...
+     * @return array
+     */
+    public function getConfig()
+    {
+        $listOfActivePaymentMethods = $this->_paymentLists->getActiveList($this->_storeManager->getStore()->getId());
+        foreach ($listOfActivePaymentMethods as $method) {
+            if ($method->getCode() === OmiseInstallmentConfig::CODE) {
+                return [
+                    'installment_config' => [
+                        OmiseInstallmentConfig::KBANK => $this->_scopeConfig->getValue('installment_config/' . OmiseInstallmentConfig::KBANK . '/active', \Magento\Store\Model\ScopeInterface::SCOPE_STORE),
```
jonrandy: Wouldn't putting these in an array in `OmiseInstallmentConfig` make it easier to extend (add new banks)? This function feels a little clunky

etc/adminhtml/system.xml
```diff
@@ -34,10 +35,11 @@
                 </field>
                 <field id="webhook" translate="label" type="label" sortOrder="70" showInDefault="1" showInWebsite="1" showInStore="1">
                     <label>Webhook endpoint</label>
-                    <comment><![CDATA[To enable <a href="https://www.omise.co/api-webhooks">WebHooks</a> feature, you must copy the above url to setup an endpoint at <a href="https://dashboard.omise.co/test/webhooks/edit"><strong>Omise dashboard</strong></a> <em>(HTTPS only)</em>.]]></comment>
+                    <comment>
+                        <![CDATA[To enable <a href="https://www.omise.co/api-webhooks">WebHooks</a> feature, you must copy the above url to setup an endpoint at <a href="https://dashboard.omise.co/test/webhooks/edit"><strong>Omise dashboard</strong></a><em>(HTTPS only)</em>.]]>
```
jonrandy: 'To enable **the** WebHooks feature... '

i18n/th_TH.csv
```diff
@@ -21,3 +21,4 @@
 "Payment rejected by card issuer.", "รายการถูกปฏิเสธโดยธนาคารผู้ออกบัตร"
 "Card stolen or lost.", "บัตรหายหรือถูกขโมย"
 "Payer did not take action before charge expiration.", "ผู้ซื้อไม่ทำรายการภายในเวลาที่กำหนด"
+"There are no any installment options available", "ไม่มีตัวเลือกในการผ่อนชำระ"
```
jonrandy: `"There are no installment options available"`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -69,6 +69,29 @@ define(
...
             },
 
+            /**
+             * Format Price
+             * 
+             * @param {string} option Name of option to be verified
+             * @return {boolean}
+             */
+            isInstallmentOptionActive: function (option) {
+                return parseInt(window.checkoutConfig.installment_config[option]);
```
jonrandy: Should be converting to a boolean, not int?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -69,6 +69,29 @@ define(
...
+            },
+
+            /**
+             * Format Price
+             * 
+             * @param {string} option Name of option to be verified
+             * @return {boolean}
+             */
+            isAnyInstallmentOptionActive: function () {
+                return parseInt(window.checkoutConfig.installment_config['omise_offsite_installment_kbank'])
```
jonrandy: A lot of this feels very rigid and inflexible. Would it not be better to change to a more flexible approach approach given that we will have the instalment options coming from Capabilities? At least moving to an architecture that is some way towards accommodating this?

view/frontend/web/template/payment/offsite-installment-form.html
jonrandy: `There are no instalment options available`

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
jacstn: we need to move from capabilities, all coming changes will be hardcoded.

jonrandy: We absolutely do not want to move from capabilities. It tells us what the merchant's account can do. We don't want to be offering them payment instalment options not supported by their account. The best way to go for now is to pull the available instalment options from Capabilities (they are already there), and create some means to switch those available ones on/off from admin in Magento - as has been requested. We should implement this in such a way that the mechanism for switching off individual instalment methods can either be switched off easily (as the switching on/off may eventually move to the Omise dashboard side), or architected in such a way that it would be easy to change the manner in which they are switch on/off (i.e. entirely decouple the means of configuring this from the way the switching on/off is implemented)

jacstn: well.. it will be hard coded, people does not have any other companies, like new company that is never listed here.

also, new pr that is coming will analyze capabilities and what installment options are available,
this pr only disable company.

jacstn: that is correct,

installment terms that are theoretically flexible for each customer, will be taken from capabilities,

that is different pr coming.

this pr only disable/enable company.

jacstn: but i get your point.. I went way tooooo far with saying EVERYTHING WILL BE HARDCODED

Model/Config/Installment.php
jacstn: my code editor does not like that.. but ok

PR #223, opened 2019-09-10
Refactoring JS View Rendering code

jacstn:
#### 1. Objective

This PR reduces code in JS view renderings.

#### 2. Description of change

Added JS file `payment-tools.js` with functions that are used in most other rendering files.
Use the new file's functionality and removed unused code.

#### 3. Quality assurance

- **Omise plugin version**: Omise-Magento 2.9.
- **PHP version**: 7.0.16.

**✏️ Details:**

Make successful payments with all payment methods.

Observe if minimum values are requested (`3000` `THB` for installment payment method),
All offsite methods and tesco method should be available only for `THB` currency configurations.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A


view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
```diff
@@ -5,15 +5,17 @@ define(
...
     ],
     function (
         ko,
         Component,
         fullScreenLoader,
         redirectOnSuccessAction,
         quote,
-        errorProcessor
+        errorProcessor,
+        paymentTools,
```
jonrandy: Would most of this refactoring be better achieved using 'mixins'? - would remove the need for all the single line stub functions that just call the same function elsewhere. Also removes the need for `paymentTools` to be passed in.

https://erfanimani.com/three-ways-to-modify-magento-2-javascript

view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
```diff
@@ -39,7 +41,7 @@ define(
              * @return {boolean}
              */
             isActive: function() {
-                return this.getOrderCurrency().toLowerCase() === 'thb' && this.getStoreCurrency().toLowerCase() === 'thb';
+                return paymentTools.getOrderCurrency().toLowerCase() === 'thb' && paymentTools.getStoreCurrency().toLowerCase() === 'thb';
```
jonrandy: Are these currencies always used in lowercase? Would it be better to move the case transformation to the currency functions themselves?

view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
jacstn: how would you do it?
I might misunderstand it, but I see `mixins` being used for other stuff, we are reducing code instead of replacing magento - modules with ours.

jonrandy: Well, we have 2 options really:

a) Make a generic payment method renderer module which contains all the common 'methods', then in the individual modules for each payment method, we would bring in that module, and return a modified verison of it (adding the stuff specific to that payment method). You'd do this by modifiying the object directly (prototypes, adding methods etc.)

b) The other way is to do similar to the above, but to use the method already provided in Magento - 'mixins'. Create a mixin with all the common functionality, and mix it in to small modules specific to each individual payment type

jonrandy: After reviewing again, it really does seem we should move to some kind of parent-child arrangement

> > The `getCode` function in each of the method renderers is called repeatedly and it just returns a string. Is there any reason why this code cannot just be a constant / property?
> 
> it needs to be a function if you want to access it from `HTML`

This is incorrect

jonrandy:
> > The `getCode` function in each of the method renderers is called repeatedly and it just returns a string. Is there any reason why this code cannot just be a constant / property?
> 
> it needs to be a function if you want to access it from `HTML`

This is incorrect

jonrandy:
> > > > The `isSandboxOn` method appears in multiple method renderers and is identical in each. Should probably be in a parent of some kind
> > > 
> > > 
> > > I was thinking about that, but it needs to be defined as part of a component of each JS renderer, to be accessed from HTML
> > 
> > 
> > This is where having a base component and extending that in each individual rendering component would be useful. There are examples of this in some of the links I provided before
> 
> I don't think so... we most probably still would have to use the same functions in component in each JS file, to be accessed from HTML

Definitely not correct as we have mechanisms available to extend a base component in one file, making functions in there available across all renderers

jacstn:
> > > The `isSandboxOn` method appears in multiple method renderers and is identical in each. Should probably be in a parent of some kind
> > 
> > 
> > I was thinking about that, but it needs to be defined as part of a component of each JS renderer, to be accessed from HTML
> 
> This is where having a base component and extending that in each individual rendering component would be useful. There are examples of this in some of the links I provided before

@jonrandy do you have any example of real Magento 2 existing module in current framework that would use your architecture proposal?

jacstn:
> > > > > The `isSandboxOn` method appears in multiple method renderers and is identical in each. Should probably be in a parent of some kind
> > > > 
> > > > 
> > > > I was thinking about that, but it needs to be defined as part of a component of each JS renderer, to be accessed from HTML
> > > 
> > > 
> > > This is where having a base component and extending that in each individual rendering component would be useful. There are examples of this in some of the links I provided before
> > 
> > 
> > I don't think so... we most probably still would have to use the same functions in component in each JS file, to be accessed from HTML
> 
> Definitely not correct as we have mechanisms available to extend a base component in one file, making functions in there available across all renderers

I think we misunderstand each other, could you please submit some proposal of your idea?
Give real example on existing code? you can do checkout from this branch.

jonrandy:
> > `isCustomerHasCard` is a bizarre function name - `customerHasCard` would be better
> 
> this is a general convention of this file,
> functions that return true/false starts with `is` except validate, and that is fine.. I guess.
> name of function despite bit broken English says exactly what it does.

`isCustomerWithSavedCard`

jacstn:
> isCustomerWithSavedCard

TBH i prefere what was developed in the beginning.

I think way better is `isCustomerHasCard` then `isCustomerWithSavedCard`
even if not that proper english
@guzzilar any thoughts?

jonrandy:
> > isCustomerWithSavedCard
> 
> TBH i prefere what was developed in the beginning.
> 
> I think way better is `isCustomerHasCard` then `isCustomerWithSavedCard`
> even if not that proper english
> @guzzilar any thoughts?

The name should **clearly** communicate the purpose of the function. Any native English speaking developer coming to this new will trip over this name and have to re-read it at least once to discern the function's purpose

view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
jacstn: thanks @jonrandy, you are welcome to submit your proposal.

jacstn:
> > > isCustomerWithSavedCard
> > 
> > 
> > TBH i prefere what was developed in the beginning.
> > I think way better is `isCustomerHasCard` then `isCustomerWithSavedCard`
> > even if not that proper english
> > @guzzilar any thoughts?
> 
> The name should **clearly** communicate the purpose of the function. Any native English speaking developer coming to this new will trip over this name and have to re-read it at least once to discern the function's purpose


It might be only me.. but my feeling is that for all non-native English speaking developer, the existing name is better, that is why I ask @guzzilar who I think introduced this name.

guzzilar:
✋ hi, the alternative solution has been provided at #225. Should we close this one? 😁

PR #225, opened 2019-09-30
Refactors JS Payment method renderers

jonrandy:
#### 1. Objective

Refactor the payment method renderers to streamline the code and ease maintenance + addition of future payment methods

#### 2. Description of change

Fairly heavy refactoring of the method renderer modules so they are composed from blocks of code that are re-used across different modules. Also some general cleanup

#### 3. Quality assurance

Re-tested all payment methods (Magento 2.3.2)

#### 4. Impact of the change

JS method renderer code now considerably smaller and hopefully easier to maintain

#### 5. Priority of change

Normal


view/frontend/web/js/view/payment/omise.js
```diff
@@ -9,47 +9,21 @@ define(
...
-            }
-        );
+        const METHOD_RENDERERS = [
+            'cc',
+            'offsite-internetbanking',
+            'offsite-alipay',
+            'offsite-installment',
+            'offsite-truemoney',
+            'offline-tesco'
+        ];
```
guzzilar: maybe it could go with `omise-` prefixes and it should be just fine for 👇 ?
```js
rendererList.push({
    type: rendererName.replace(/-/g, '_'),
    component: 'Omise_Payment/js/view/payment/method-renderer/' + rendererName + '-method'
});  
```

or you prefer to reduce the repeating `omise` prefix here as well?

view/frontend/web/js/view/payment/omise-offsite-method-renderer.js
```diff
@@ -0,0 +1,69 @@
...
+        'Omise_Payment/js/view/payment/omise-offline-method-renderer',
+        'Magento_Checkout/js/model/error-processor',
+        'Magento_Checkout/js/model/full-screen-loader',
+        'Magento_Checkout/js/model/url-builder',
+        'mage/storage',
+        'jquery'
+    ],
+    function (
+        Base,
+        errorProcessor,
```
guzzilar: Do we need these 2 libs here? (`errorProcessor`, `fullScreenLoader`). Seems there is no place using it in this file.

view/frontend/web/js/view/payment/omise-base-method-renderer.js
```diff
@@ -0,0 +1,59 @@
...
+    ],
+    function (
+        errorProcessor,
+        fullScreenLoader
+    ) {
+        'use strict';
+
+        return {
+
+            OFFSITE_RETURN_URL: '/orders/:order_id/omise-offsite',
```
guzzilar: I cannot remember what it is exactly but it seems not a return url (maybe API Service Provider or something).
@jacstn what name do you think it is? 😂

view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
```diff
@@ -77,23 +35,17 @@ define(
              */
             placeOrder: function(data, event) {
                 var self = this;
+                var failHandler = this.buildFailHandler(self);
```
guzzilar: In a case where we encapsulate everything inside `this.buildFailHandler` now. 
Do we still need `var self = this` and all the `self` variables here?

view/frontend/web/js/view/payment/omise-offsite-method-renderer.js
```diff
@@ -0,0 +1,69 @@
+define(
+    [
+        'Omise_Payment/js/view/payment/omise-offline-method-renderer',
```
guzzilar: Just feel weird that offsite method renderer is extending its base from offline renderer.
Is there any reason why we shouldn't extend from `omise-base-method-renderer.js`? (besides sharing `isActive` function)

guzzilar: If just only because it is sharing `isActive` behaviour then I'd like to propose to move 
`getOrderCurrency` and `getStoreCurrency` to `omise-base-method-renderer.js` then let's have `isActive` both on `omise-offline-method-renderer.js` and `omise-offsite-method-renderer.js `. 

Although the code will be duplicated but I think it more makes sense that `offsite` and `offline` should be extending its capability from `base-method-renderer`.

view/frontend/web/js/view/payment/omise-base-method-renderer.js
guzzilar: Actually seeing this url is being used only to build a `serviceUrl` and now wonder if it would be better if we just have a function to generate this url? Could be something like 👇 
```js
buildServiceUrl: function (order_id) {
    return urlBuilder.createUrl(
        '/orders/:order_id/omise-offsite',
        { order_id: order_id }
    );
}
```

Although, I am still not sure if we need this – just I forgot what it was for.

view/frontend/web/js/view/payment/omise-offsite-method-renderer.js
jonrandy: Well spotted - I thought I got all these 👍

jonrandy: Well spotted - I thought I got all these 👍

view/frontend/web/js/view/payment/method-renderer/omise-offline-tesco-method.js
jonrandy: 👍

view/frontend/web/js/view/payment/omise.js
jonrandy: I didn't want to start doing major renaming when I knew Jacek has another payment method incoming. I agree, it makes sense though

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
```diff
@@ -265,7 +217,7 @@ define(
                     return false;
                 }
 
-                this.generateTokenAndPerformPlaceOrderAction(data);
+                this.generateTokenAndPerformPlaceOrderAction();
```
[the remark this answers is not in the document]
jonrandy: I left Jacek's function names alone in this refactor, but I agree this function name is a bit on the wordy side

view/frontend/web/js/view/payment/omise-offsite-method-renderer.js
jonrandy: Offsite extends from offline, Offline extends from `omise-base-method-renderer.js` - was the best way to keep code only where needed

guzzilar: Programmatically, I would agree on this as there is no single code that has to be duplicated.

However, talking about an intention of the code as a reader perspective.
Reading this one and it could be implied that `Offsite` is either subset of `Offline` or have something related to `Offline` method while in fact, it just happens to have a similar behaviour on `isActive` function. Nothing more or less.

jonrandy: Names could be better, could name 'omise-offline-method-renderer.js' as 'active-for-thb-method-renderer.js` or something like that. Then it's like we are combining traits to make up the final renderers

guzzilar: Alternative idea, how about we move the whole functions in `offline-method-renderer.js` to `base`?

```js
/**
 * Is method available to display
 *
 * @return {boolean}
 */
isActive: function() {
    return this.getOrderCurrency() === 'thb' && this.getStoreCurrency() === 'thb';
},

/**
 * Get order currency
 *
 * @return {string}
 */
getOrderCurrency: function () {
    return window.checkoutConfig.quoteData.quote_currency_code.toLowerCase();
},

/**
 * Get store currency
 *
 * @return {string}
 */
getStoreCurrency: function () {
    return window.checkoutConfig.quoteData.store_currency_code.toLowerCase();
}
```

This whole part.
Then, `omise-cc-method.js` can have this 
```js
isActive: function() {
    return true;
}
```

Also we can move `placeOrder` from `omise-offline-tesco-method.js` to `offline-method-renderer.js`

jonrandy: Seems like not a great idea as it would imply that all new payment methods by default are only for THB

jonrandy: We could add a var something like `restrictedToCurrencies` which could contain an array of valid currencies for the payment method. Then we could have a single `isActive` based on this var (returns `true` if no restrictions)

guzzilar: another one more alternative idea, how about this:
Move `getOrderCurrency`, `getStoreCurrency` from `offline-method-renderer.js` to `base-method-render.js` 
```js
// omise-base-method-renderer.js
define(
    [
        'Magento_Checkout/js/model/error-processor',
        'Magento_Checkout/js/model/full-screen-loader'
    ],
    function (
        errorProcessor,
        fullScreenLoader
    ) {
        'use strict';

        return {
            // ... the same code

            /**
             * Get order currency
             *
             * @return {string}
             */
            getOrderCurrency: function () {
                return window.checkoutConfig.quoteData.quote_currency_code.toLowerCase();
            },

            /**
             * Get store currency
             *
             * @return {string}
             */
            getStoreCurrency: function () {
                return window.checkoutConfig.quoteData.store_currency_code.toLowerCase();
            }
        };
    }
);
```

Then `offline-method-renderer.js` just
```js
define(
    [
        'Omise_Payment/js/view/payment/omise-base-method-renderer'
    ],
    function (
        Base
    ) {
        'use strict';

        return Object.assign({}, Base, {

            /**
             * Is method available to display
             *
             * @return {boolean}
             */
            isActive: function() {
                // ...
            },

            /**
             * Hook the placeOrder function.
             * Original source: placeOrder(data, event); @ module-checkout/view/frontend/web/js/view/payment/default.js
             *
             * @return {boolean}
             */
            placeOrder: function(data, event) {
                // ...
            }
        });
    }
);
```

The same as `offsite-method-renderer.js`
```js
define(
    [
        'Omise_Payment/js/view/payment/omise-base-method-renderer',
        'Magento_Checkout/js/model/url-builder',
        'mage/storage',
        'jquery'
    ],
    function (
        Base,
        urlBuilder,
        storage,
        $
    ) {
        'use strict';

        return Object.assign({}, Base, {

            /**
             * Is method available to display
             *
             * @return {boolean}
             */
            isActive: function() {
                // ...
            },

            /**
             * Hook the placeOrder function.
             * Original source: placeOrder(data, event); @ module-checkout/view/frontend/web/js/view/payment/default.js
             *
             * @return {boolean}
             */
            placeOrder: function (data, event) {
                // ...
            }
        });
    }
);
```

guzzilar: Here we take `isActive` and `placeOrder` as sort of `interface` functions that all `payment-flow` should provide.
In this case, `offline`, `offsite`, `creditcard`

guzzilar: > We could add a var something like restrictedToCurrencies which could contain an array of valid currencies for the payment method. Then we could have a single isActive based on this var (returns true if no restrictions)

Personally I think we don't really need to reduce down line of code to the point that it's super lean but don't feel right. But I think this would work too, if you prefer 👍

view/frontend/web/js/view/payment/omise-base-method-renderer.js
jonrandy: Can we agree a name here? @jacstn

Then we can move this to a general method in `base`

jonrandy: `getMagentoReturnUrl`?

jonrandy: Done! Want to check again?

jacstn:
@guzzilar are you happy with new changes?

guzzilar:
I'm fine by the change. However, still having a feeling that `OFFSITE_RETURN_URL` constant naming is a bit not the right name.

Just checked its original commit message where it was introduced. 
```
Implement RESTFul API to retrieve a payment detail after place an order.    
To retrieve an authorize_uri back from charge result.
```

It's more like a RESTFul API's URI that will be called immediately right after an order has been placed. However, I have also no idea what it should be named. 😂

PR #226, opened 2019-11-11
Japaniese translations of Magento 2 Plugin

jacstn:
#### 1. Objective

Translate plugin to the Japanese language translation

#### 2. Description of change

Added CSV

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

Switch translation language to Japanese, check if all translations are displayed correctly.

#### 4. Impact of the change

N/A

Note: Please provide a screenshot if your changed impact to UI.

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A


i18n/jp_JP.csv
```diff
@@ -0,0 +1,15 @@
+"Convenience Store", "コンビニエンスストア"
```
guzzilar: Not sure if we should use `コンビニ決済` instead.
`コンビニエンスストア` is a correct translation but probably not for a context of the payment setting page.
Can we somehow double check this a bit?

kotaura-omise: Thanks for checking! `コンビニ決済` is okay. The actual translation for Convenience store is `コンビニエンスストア`, but this context refers to payment method. We can pay by `コンビニ決済` at the `コンビニエンスストア` so bottom line `コンビニ決済` is fine :) Thanks so much again!

PR #231, opened 2019-11-28
Citi Points Payment Implementation

jacstn:
#### 1. Objective
This PR enables Pay with Citi Points Method.

**Note** Payment is available only for customers with Thai account.
More info about Citi Points Payment is available here: https://www.omise.co/pay-with-points

#### 2. Description of change

Added Virtual and PHP classes to support Citi Points Payment.
Added `javascript` and `HTML` UI rendering files.

Admin Panel:
image

Checkout page:
image

Checkout Success page
image

#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: Magento CE 2.5.
- **Omise plugin version**: Omise-Magento 2.8
- **PHP version**: 7.0.29.

**✏️ Details:**
_You need to use keys from Thai Omise account._

Turn on a `Citi Points Payment` in the admin panel, check if payment is available on the checkout page.

Turn off a `Citi Points Payment` method on the admin panel, the option should not be listed on the checkout page anymore.

Make an order and successful payment with the `Citi Points Payment` method.

Verify if the proper amount and payment method are displayed in [Omise Dashboard](https://dashboard.omise.co)

#### 4. Impact of the change

The new Payment Method is available for Thai merchants using Magento 2.

#### 5. Priority of change

Normal

#### 6. Additional Notes

none


Controller/Callback/Offsite.php
```diff
@@ -84,7 +85,7 @@ public function execute()
         }
         
         $paymentMethod = $payment->getMethod();
-        if (! in_array($paymentMethod, [Alipay::CODE, Internetbanking::CODE, Installment::CODE, Truemoney::CODE])) {
+        if (! in_array($paymentMethod, [Alipay::CODE, Internetbanking::CODE, Installment::CODE, Truemoney::CODE, Citipoints::CODE])) {
```
jonrandy: Feels like this array should be named and defined elsewhere - maybe `validPaymentMethods` or something

jonrandy: Also, this `execute` function is HUGE - maybe we should break it up?

Controller/Callback/Offsite.php
```diff
@@ -143,6 +144,9 @@ public function execute()
                     case Truemoney::CODE:
                         $dispPaymentMethod = "True Money";
                         break;
+                    case Citipoints::CODE:
+                        $dispPaymentMethod = "Citi Points";
```
jonrandy: Is this hard coded English?

etc/adminhtml/system.xml
```diff
@@ -94,6 +94,21 @@
                         <config_path>payment/omise_offsite_alipay/title</config_path>
                     </field>
                 </group>
+                <group id="omise_offsite_citipoints" translate="label" sortOrder="160" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable Citi Points Solution</label>
+                    <comment>Enables customers to pay using Citi Points method</comment>
+                    <field id="active" translate="label comment" type="select" sortOrder="170" showInDefault="1" showInWebsite="1" showInStore="1">
+                        <label>Enable/Disable</label>
```
jonrandy: The label 'enable/disable' makes no sense with the options 'yes' and 'no'. 'Enabled?' would be a better choice

jacstn: thanks... so what better messages would you recommend.
just notice that it is all payment methods have similar texts, so I will change in separated PR all of them.

I would leave enable/disable but change title. or what do you think?

Controller/Callback/Offsite.php
jacstn: how would you break it up?

I don't hate that array defined here, what @guzzilar thinks about that?

Controller/Callback/Offsite.php
jacstn: what is the question?

Controller/Callback/Offsite.php
jacstn: also:
Any refactoring changes that are not consistent with any file, might be done in a separated file.

etc/adminhtml/system.xml
jacstn: I changed this texts here: #232

Controller/Callback/Offsite.php
jacstn: those texts are added to the order log in the admin panel.

guzzilar:
I haven't tested the code but just quick thought about naming.
"Citipoints" seems a bit too specific. Reading from Omise document where `source.type` is `points_citi`. I'm guessing that this is to prepare for the (unknown coming) extending feature, the same idea as Internet Banking payment ( `internet_banking_***` – in this case, `points_***` ).

Maybe name the class as `Points` would be more flexible in case if it's needed to be extend?

Controller/Callback/Offsite.php
guzzilar: IIRC, I think this topic has been discussed several times and I agreed that this method is getting big and would be nice if we can lean it down a bit to avoid any potential one error break all payments.

If you ask me, as I don't have a particular idea on how to do it, so I'm ok with what it is for now. Maybe I'll also submit an alternative refactoring idea next year on Jan or Feb 😉

etc/adminhtml/system.xml
```diff
@@ -94,6 +94,21 @@
                         <config_path>payment/omise_offsite_alipay/title</config_path>
                     </field>
                 </group>
+                <group id="omise_offsite_citipoints" translate="label" sortOrder="160" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable Citi Points Solution</label>
```
guzzilar: Maybe Citi Points solution is a bit ambiguous? As it is not our marketing term for this feature. 
Maybe "Pay with City Points Solution"?

view/frontend/web/js/view/payment/method-renderer/omise-offsite-citipoints-method.js
```diff
@@ -0,0 +1,27 @@
...
+                template: 'Omise_Payment/payment/offsite-citipoints-form'
+            },
+
+            isPlaceOrderActionAllowed: ko.observable(quote.billingAddress() != null),
+
+            code: 'omise_offsite_citipoints',
+            restrictedToCurrencies: ['thb']
+        });
+    }
+);
```
guzzilar: lovely result from refactoring the JS code! Thanks @jacstn @jonrandy 😍

jacstn:
> I haven't tested the code but just quick thought about naming.
> "Citipoints" seems a bit too specific. Reading from Omise document where `source.type` is `points_citi`. I'm guessing that part of the naming idea is to prepare for the (unknown coming) extending feature, the same idea as Internet Banking payment ( `internet_banking_***` – in this case, `points_***` ).
> 
> Maybe name the class as `Points` would be more flexible in case if it's needed to be extend?

@guzzilar do you have any info about that?
Yes, I treat it as specific as ie. `alipay`,
I am not going to change that at the moment if we have no specific info.
As we have very limited time today, let's focus on key stuff in this review.

etc/adminhtml/system.xml
jacstn: Agree that is way better name

Controller/Callback/Offsite.php
jonrandy: Do they get translated?

etc/adminhtml/system.xml
jonrandy: Removing 'Solution' is probably better

guzzilar:
> @guzzilar do you have any info about that?
Yes, I treat it as specific as ie. alipay,
I am not going to change that at the moment if we have no specific info.
As we have very limited time today, let's focus on key stuff in this review.

@jacstn No, I don't have info, but reading the intention of the source.type and it seems so.

> As we have very limited time today, let's focus on key stuff in this review.

We have time, I just don't have it haha. 
How about we carry this topic to discuss with the payment team for their opinions before deciding that this one is an unnecessary point one, just to unclear the myst? : )

Controller/Callback/Offsite.php
jacstn: no, no necessary.

etc/adminhtml/system.xml
jacstn: done in #232

jonrandy: Apparently "Citi points" is not correct - according to Robin, the correct name is "City Pay with Points"

jacstn:
I corrected it @guzzilar citi points to points citi

jonrandy:
TBH I still have strong concern about the huge method mentioned [here](https://github.com/omise/omise-magento/pull/231#discussion_r356980966). The fact that Nam said it has been mentioned numerous times before suggested that it has also been ignored numerous times before - reducing confidence that will ever get resolved. At the start of any new feature we should consider if what we are about to do is really the best way to achieve it or could it be better achieved by giving ourselves a more sensible, maintainable base to build the new work from. If we don't do this as we go along, we end up with monoliths that do not easily lend themselves to refactoring and all associated problems are compounded.

jacstn:
> TBH I still have strong concern about the huge method mentioned [here](https://github.com/omise/omise-magento/pull/231#discussion_r356980966). The fact that Nam said it has been mentioned numerous times before suggested that it has also been ignored numerous times before - reducing confidence that will ever get resolved. At the start of any new feature we should consider if what we are about to do is really the best way to achieve it or could it be better achieved by giving ourselves a more sensible, maintainable base to build the new work from. If we don't do this as we go along, we end up with monoliths that do not easily lend themselves to refactoring and all associated problems are compounded.

thanks @jonrandy, I am comfortable with that as it is now, since there is only `switch` and `in_array` added.
I am comfortable to add more payment methods here, there are no any specific functions added, only `case` into `switch`, not a biggie for me.

jonrandy:
Sorry, but I cannot in good conscience approve the PR. Two reviewers have now expressed concern over the way it is structured and have stressed the need for refactoring. We cannot go on simply ignoring stuff like this

jacstn:
area you are concerned is related to bigger picture, than this PR.
so could you explain what you cannot agree on @jonrandy ?

even if any changes will be submitted regarding your concerns, they will be compleately in another PR.

jonrandy:
What I am concerned with is directly related to this PR as this PR has been built on top of code that concerns were expressed about before, without anything being done about those concerns. You have also insinuated that you possibly intend to continue ignoring these concerns and building on top of this code.

You are actively ignoring the concerns of your reviewers, rendering the process of code review all but useless.

We have to draw a line somewhere

jonrandy:
Wow. Really?

PR #234, opened 2019-12-20
Refactoring duplicated code in Model and Callback controller classes

mayurkathale:
#### 1. Objective
Refactoring to remove duplicated code in all payment methods.

#### 2. Description of change
- Removed duplicate code using Base classes.
- Longer functions broken into smaller.
- Added documentation

#### 3. Quality assurance
🔧 Environments:
Omise plugin version: Omise-Magento 2.1.
PHP version: 5.6.
✏️ Details:
Turn on a Pay with Citipoints in the admin panel, check if payment is available on the checkout page.

Turn off a Pay with Citipoints method on the admin panel, the option should not be listed on the checkout page anymore.

Make an order and successful payment with the Pay with Citipoints method.

Verify if the proper amount and payment method are displayed in Omise Dashboard

#### 4. Impact of the change

All Payment Method available for Thai merchants should work normally Magento 1

#### 5. Priority of change

Normal

#### 6. Additional Notes

NA

### DIFF
--- a/src/app/code/community/Omise/Gateway/Controller/Base.php
+++ b/src/app/code/community/Omise/Gateway/Controller/Base.php
@@ -4,11 +4,6 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
     /**
      * @var string
      */
-    protected $title;
-
-    /**
-     * @var string
-     */
     protected $message;
 
     /**
@@ -29,22 +24,6 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
     /**
      * @return string
      */
-    public function getTitle()
-    {
-        return $this->title;
-    }
-
-    /**
-     * @param string $title
-     */
-    public function setTitle($title)
-    {
-        $this->title = $title;
-    }
-
-    /**
-     * @return string
-     */
     public function getMessage()
     {
         return $this->message;
@@ -59,29 +38,13 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
     }
 
     /**
-     * @return string
-     */
-    public function getAwaitingOrderStatus()
-    {
-        return $this->awaitingOrderStatus;
-    }
-
-    /**
-     * @param string $awaitingOrderStatus
-     */
-    public function setAwaitingOrderStatus($awaitingOrderStatus)
-    {
-        $this->awaitingOrderStatus = $awaitingOrderStatus;
-    }
-
-    /**
      * constructor
      */
     protected function _construct()
     {
         $omise = Mage::getModel('omise_gateway/omise');
         $omise->initNecessaryConstant();
-        $this->setAwaitingOrderStatus(Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW);
+        $this->awaitingOrderStatus = Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW;
     }
 
     /**
@@ -103,8 +66,8 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
             $this->order = $this->_getOrder();
         if (! $payment = $this->order->getPayment()) {
             Mage::getSingleton('core/session')->addError(
-                $this->__($this->getTitle().' validation failed, we cannot retrieve your payment information. 
-                Please contact our support team to confirm the payment.'
+                $this->__('%s validation failed, we cannot retrieve your payment information. 
+                Please contact our support team to confirm the payment.', self::PAYMENT_TITLE
                 )
             );
             $this->_redirect('checkout/cart');
@@ -155,11 +118,11 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
             $this->getAwaitingOrderStatus(),
             Mage::helper('omise_gateway')->__(
                 (empty($this->getMessage()))
-                    ? 'The payment is in progress.<br/>Due to the way '.$this->getTitle().' works, this might take 
+                    ? 'The payment is in progress.<br/>Due to the way %s works, this might take 
                         a few seconds or up to an hour. Please click "Accept" or "Deny" to complete the payment manually 
                         once the result has been updated (you can check at Omise Dashboard).'
                     : $this->getMessage(),
-                $this->getTitle()
+                self::PAYMENT_TITLE
             )
         );
         if($this->getAwaitingOrderStatus() != Mage_Sales_Model_Order::STATE_PROCESSING) {
--- a/src/app/code/community/Omise/Gateway/Controller/Base.php
+++ b/src/app/code/community/Omise/Gateway/Controller/Base.php
@@ -19,7 +19,7 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
     /**
      * @var string
      */
-    protected $awaitingOrderStatus;
+    protected $awaitingOrderStatus = Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW;
 
     /**
      * @return string
@@ -44,7 +44,6 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
     {
         $omise = Mage::getModel('omise_gateway/omise');
         $omise->initNecessaryConstant();
-        $this->awaitingOrderStatus = Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW;
     }
 
     /**
@@ -52,7 +51,7 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
      */
     protected function _getOrder()
     {
-        $id = $this->getRequest()->getParam('order_id') ? $this->getRequest()->getParam('order_id') : null;
+        $id = $this->getRequest()->getParam('order_id') ?: null;
 
         return Mage::getModel('omise_gateway/order')->getOrder($id);
     }
--- a/src/app/code/community/Omise/Gateway/Controller/Base.php
+++ b/src/app/code/community/Omise/Gateway/Controller/Base.php
@@ -66,7 +66,7 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
             $this->order = $this->_getOrder();
         if (! $payment = $this->order->getPayment()) {
             Mage::getSingleton('core/session')->addError(
-                $this->__('%s validation failed, we cannot retrieve your payment information. 
+                $this->__('%1 validation failed, we cannot retrieve your payment information. 
                 Please contact our support team to confirm the payment.', self::PAYMENT_TITLE
                 )
             );
@@ -101,7 +101,7 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
         }
         $order->markAsFailed(
             $this->transId,
-            $this->__('The payment was invalid, %s (%s)', $charge->failure_message, $charge->failure_code)
+            $this->__('The payment was invalid, %1 (%2)', $charge->failure_message, $charge->failure_code)
         );
         return $this->_redirect('checkout/cart');
     }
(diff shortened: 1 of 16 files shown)


src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,194 @@
 <?php
 abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_Action
 {
+    /**
+     * @var string
+     */
+    protected $title;
```
jonrandy: Unless we need to change this (and many others below) at any point, I think we'd be better off just having them as class constants (or maybe public variables). The getters and setters certainly aren't needed - they are just adding unnecessary overhead

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,194 @@
...
+    /**
+     * @return Mage_Core_Controller_Varien_Action|void
+     * @throws Mage_Core_Exception
+     */
+    protected function validate() {
+        if(!isset($this->order))
+            $this->order = $this->_getOrder();
+        if (! $payment = $this->order->getPayment()) {
+            Mage::getSingleton('core/session')->addError(
+                $this->__($this->getTitle().' validation failed, we cannot retrieve your payment information.
```
jonrandy: If there are strings to be inserted to translatable text, we need to avoid having those strings as part of the translation - otherwise we need to edit all the translations if (in this case) the title changes. The translation function of Magento has the facility to insert strings within a translation as follows:

`$this->__('%1 validation failed... %2...', 'string to replace %1', 'string to replace %2', ...)`

In this case though, we may also want to consider translating the title separately, so we would have:

`$this->__('%1 validation failed, we cannot retrieve your payment information. Please contact our support team to confirm the payment.', $this->__($this->getTitle())`

Although `getTitle` can probably also go (see above comment)

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,194 @@
...
+     * @param Omise_Gateway_Model_Order $order
+     * @return Mage_Core_Controller_Varien_Action
+     */
+    protected function paymentAwaiting($order) {
+        $order->markAsAwaitPayment(
+            $this->transId,
+            $this->getAwaitingOrderStatus(),
+            Mage::helper('omise_gateway')->__(
+                (empty($this->getMessage()))
+                    ? 'The payment is in progress.<br/>Due to the way '.$this->getTitle().' works, this might take
```
jonrandy: Issue again here with usage of translation function.

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,194 @@
...
+     * @return Mage_Core_Controller_Varien_Action
+     */
+    protected function paymentAwaiting($order) {
+        $order->markAsAwaitPayment(
+            $this->transId,
+            $this->getAwaitingOrderStatus(),
+            Mage::helper('omise_gateway')->__(
+                (empty($this->getMessage()))
+                    ? 'The payment is in progress.<br/>Due to the way '.$this->getTitle().' works, this might take 
+                        a few seconds or up to an hour. Please click "Accept" or "Deny" to complete the payment manually
```
jonrandy: Probably needs rewording - "Deny" is not going to complete the payment. Maybe `Please click "Accept" or "Deny" to complete the transaction process"`

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,194 @@
...
+    protected function checkPaymentStatusAndUpdateOrder($charge, $order) {
+        if ($charge->isAwaitPayment()) {
+            return $this->paymentAwaiting($order);
+        }
+        if ($charge->isSuccessful()) {
+            return $this->paymentSuccessful($order);
+        }
+        $order->markAsFailed(
+            $this->transId,
+            $this->__('The payment was invalid, %s (%s)', $charge->failure_message, $charge->failure_code)
```
jonrandy: This may need modification if inserted strings are not automatically translated

src/app/code/community/Omise/Gateway/Model/Payment/Base/Payment.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * @var string
+     */
+    protected $_callbackUrl;
+
+    /**
+     * @param string $code
+     */
+    public function setCode($code)
```
jonrandy: Again, since a lot of these don't get changed, we would be better off with class vars or constants, and removing the getters and setters which aren't needed

src/app/code/community/Omise/Gateway/Model/Payment/Base/Payment.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * @return string
+     */
+    public function getCallbackUrl()
+    {
+        return $this->_callbackUrl;
+    }
+
+    public function __construct()
```
jonrandy: We can remove this - if it isn't defined, the parent constructor will be called automatically

src/app/code/community/Omise/Gateway/Model/Payment/Base/Payment.php
```diff
@@ -0,0 +1,193 @@
...
+
+            return;
+        }
+
+        $this->_suspectToBeFailed($payment);
+    }
+
+    /**
+     * Check method for processing with base currency
+     * Note that Internet Banking can only be used with Omise Thailand account and 'THB' currency
```
jonrandy: This comment is referring to Internet Banking

src/app/code/community/Omise/Gateway/Model/Payment/Base/Payment.php
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param  string $currencyCode
+     *
+     * @return boolean
+     */
+    public function canUseForCurrency($currencyCode)
+    {
+        return (strtoupper($currencyCode) === 'THB' && strtoupper(Mage::app()->getStore()->getCurrentCurrencyCode()) === 'THB');
```
jonrandy: Should probably make this look at a class variable (possibly an array) that defines the currency or currencies that a particular payment method is available for. This would make this generic `canUseForCurrency` method more flexible and remove the need to override it in as many cases

src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
```diff
@@ -1,32 +1,19 @@
 <?php
-class Omise_Gateway_Model_Payment_Creditcard extends Omise_Gateway_Model_Payment
+class Omise_Gateway_Model_Payment_Creditcard extends Omise_Gateway_Model_Payment_Base_Payment
 {
-    /**
-     * @var string
-     */
-    protected $_code = 'omise_gateway';
```
jonrandy: I think it is better to leave these and remove the getters and setters as mentioned in other comments

jonrandy: We then also remove the need for the constructor method

src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitealipayController.php
```diff
@@ -1,59 +1,14 @@
...
-            return $this->_redirect('checkout/onepage/success');
-        }
-
-        $order->markAsFailed(
-            $payment->getLastTransId(),
-            $this->__('The payment was invalid, %s (%s)', $charge->failure_message, $charge->failure_code)
-        );
-
-        return $this->_redirect('checkout/cart');
+        $this->getTitle(self::PAYMENT_TITLE);
```
jonrandy: Why would you be passing something to a getter?

src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsitecitipointsController.php
```diff
@@ -0,0 +1,14 @@
...
+class Omise_Gateway_Callback_ValidateoffsitecitipointsController extends Omise_Gateway_Controller_Base
+{
+    const PAYMENT_TITLE = 'Citi Reward Points';
+    /**
+     * @return Mage_Core_Controller_Varien_Action|void
+     * @throws Mage_Core_Exception
+     */
+    public function indexAction()
+    {
+        $this->getTitle(self::PAYMENT_TITLE);
```
jonrandy: Passing something to a getter?

src/app/code/community/Omise/Gateway/controllers/Callback/ValidateoffsiteinstallmentController.php
```diff
@@ -1,57 +1,15 @@
...
-            return $this->_redirect('checkout/onepage/success');
-        }
-
-        $order->markAsFailed(
-            $payment->getLastTransId(),
-            $this->__('The payment was invalid, %s (%s)', $charge->failure_message, $charge->failure_code)
-        );
-
-        return $this->_redirect('checkout/cart');
+        $this->getTitle(self::PAYMENT_TITLE);
```
jonrandy: Again

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,153 @@
...
+    }
+
+    /**
+     * constructor
+     */
     protected function _construct()
     {
         $omise = Mage::getModel('omise_gateway/omise');
         $omise->initNecessaryConstant();
+        $this->awaitingOrderStatus = Mage_Sales_Model_Order::STATE_PAYMENT_REVIEW;
```
jonrandy: Could we move this to the declaration?

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -1,19 +1,153 @@
...
     }
 
     /**
-     * @return \Omise_Gateway_Model_Order
+     * @return Omise_Gateway_Model_Order
      */
     protected function _getOrder()
     {
         $id = $this->getRequest()->getParam('order_id') ? $this->getRequest()->getParam('order_id') : null;
```
jonrandy: Probably better to use:
`$this->getRequest()->getParam('order_id') ?: null;`
to avoid calling all this stuff twice.

src/app/code/community/Omise/Gateway/Model/Payment/Base/Payment.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param  string $currencyCode
+     *
+     * @return boolean
+     */
+    public function canUseForCurrency($currencyCode)
+    {
+        if(isset($this->_currencies))
+            return (
+                in_array(strtoupper($currencyCode), $this->_currencies)
```
jonrandy: Wondering if it might be an idea to put the currencies array items in lower case, as Magento seems to be using lower case for currency codes and it would remove the need for these `strtoupper` conversions

src/app/code/community/Omise/Gateway/Model/Payment/Offsitecitipoints.php
```diff
@@ -0,0 +1,76 @@
...
+    protected $_canReviewPayment   = true;
+    protected $_isInitializeNeeded = true;
+
+    /**
+     * @param  Varien_Object $payment
+     * @param  float         $amount
+     *
+     * @return Omise_Gateway_Model_Api_Charge
+     */
+    public function process(Varien_Object $payment, $amount)
```
jonrandy: Looks like there might be some scope for more refactoring here since Citi and Alipay seem to have all but identical `process` methods. Maybe an intermediary sub class from the base one (Omise_Gateway_Model_Payment_SimpleOffsite_Payment?)

Thoughts @jacstn ?

src/app/code/community/Omise/Gateway/Model/Payment/Offsitecitipoints.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return parent::process($payment, $amount, $this->_type);
```
jonrandy: If we move `$this->_type` into the parent method, we can remove this method completely from this class

PR #237, opened 2020-01-23
Display warning about availability of convenience store payment method

jacstn:
#### 1. Objective

The goal of this PR is to add an explanation to the user if and why the convenience store payment method is not available.
Similar mechanisms were introduced to THB payment methods like Internet Banking or Alipay.

#### 2. Description of change

Added HTML code defining dialog displayed if the convenience store payment method is inactive.
The method is inactive if the currency of a store is different than JPY

image

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

Make store currency JPY, the message should NOT be displayed in the checkout form.
Make store currency THB, check if the message is properly displayed in the checkout form.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

N/A


jonrandy:
Surely it would be better to just not display the payment method?

jacstn:
could be.. but currency can be changed in some stores by customer.. depends on settings

jonrandy:
Yes, and if they change currency, any payment methods they can't use should disappear. Why would you show them payment methods they can't use?

jacstn:
so they know.. that payment is available if they change currency. how otherwise they know?

jonrandy:
It will appear when they change currency. The payment methods displayed should be based on the currency the customer has currently selected

jacstn:
they might not even know that that payment method is available for them. that it exists in this store.

jonrandy:
They have made a choice to select a currency. When they have done that, why on earth would they want to know what payment methods are available for other currencies other than the one they have actively selected as one they want to use?

jacstn:
also, this for a plugin to be consistent with what we have now in THB methods, if a problem will ever occur, that people will see it, it will be mostly because the merchant has not configured store properly.

jonrandy:
Thoughts @guzzilar ?

jacstn:
> They have made a choice to select a currency. When they have done that, why on earth would they want to know what payment methods are available for other currencies other than the one they have actively selected as one they want to use?

there is default currency, people might not actively make a choice of currency.

jonrandy:
They would if they intend to pay in another currency

guzzilar:
Not sure what was my opinion on this back then, but at the moment, I tend to go with a principle of "leaving things out, make it simple as possible unless someone asks to have it (and at that time, we can decide to have an option to switch thing on and off for admin)"

Considering based on an assumption (no statistic to prove) that in general people often pay by their own local currency – so, I would go with "not displaying it at all".

However, just my opinion. If you guys insist on showing it, then I can approve this one (this can work another way around too, displaying it, until someone asks loud enough to remove it out)

jonrandy:
My feeling is that we shouldn't display it at all too.

guzzilar:
Just add a little point, about how buyers can know if there is an extra payment method available if they place an order with another currency.

I think that could up to marketing strategy of that website. i.e. Marketer can simply put a logo, text, or a sentence at their checkout page or somewhere in the website saying: "Konbini payment is available only when place an order using Japanese yen"

jonrandy:
But if the user wants to pay using yen, they'll select yen, and see Konbini. I would think choosing a payment method generally comes after choosing a currency. If the customer is familiar with that payment method and want to use it, they'll already know what currency it's for, right?

guzzilar:
I think what this PR is trying to offer is for an edge case
i.e. If I can put a quick story, it might be: "A foreign guy living in Japan, buying some groceries online, maybe this website provides an English version with USD as a default currency. Then he about to use his credit card, but discovered later that he could instead pay with his Japanese Yen money (in cash), to avoid the exchange rate, so he switches to JPY currency at the last moment and using his cash."

But, I think that can simply be solved with web design (like, adding payment information somewhere on the website). Although considering the edge case above, I still think it's good idea to simply remove it and test if there are really demands on this case.

guzzilar:
However, both have its point. So, I'm fine either way (just, if you ask my thought then, the above)

jonrandy:
I agree that removing seems best

jonrandy (commented):
:)

aashishgurung:
Stale PR.

PR #238, opened 2020-01-27
Cancel the order and restore cart as soon as user cancels offsite payment transaction

mayurkathale:
#### 1. Objective

This PR addresses the issue when user cancels the payment. After cancellation payment, user returns back to Magento site on success page and cart is cleared. As order isn't placed, if user choose to create same order with different payment method, then he has to do whole process of checkout again. This PR changes the workflow by cancelling order and restoring cart after payment cancellation.

#### 2. Description of change

- Updated callback class by adding new methods.

Before PR implementation, Payment cancellation lands on success page.
Screen Shot 2020-01-27 at 15 29 50
After PR implementation, Payment cancellations restores cart items and lands on cart page.
image

#### 3. Quality assurance

**🔧 Environments:**
Omise plugin version: Omise-Magento 2.1.
PHP version: 5.6.

**✏️ Details:**
1. Add product to cart and checkout with any offsite payment method.
2. Payment success should return to Magento checkout success page.
3. Payment failure should return to Magento checkout failure page.
4. After cancellation of payment, order should get cancelled and cart should restored. Cancellation of payment should return to checkout/cart page.

#### 4. Impact of the change

User doesn't need to add items to cart after cancellation of payment.

#### 5. Priority of change

Normal

#### 6. Additional Notes

NA

### DIFF
--- a/src/app/code/community/Omise/Gateway/Controller/Base.php
+++ b/src/app/code/community/Omise/Gateway/Controller/Base.php
@@ -93,7 +93,7 @@ abstract class Omise_Gateway_Controller_Base extends Mage_Core_Controller_Front_
      */
     protected function checkPaymentStatusAndUpdateOrder($charge, $order) {
         if ($charge->isAwaitCapture() || $charge->isAwaitPayment()) {
-            return $this->paymentAwaiting($order);
+            return $this->paymentAwaiting($order, $charge);
         }
         if ($charge->isSuccessful()) {
             return $this->paymentSuccessful($order);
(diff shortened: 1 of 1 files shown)


src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -109,9 +109,12 @@ protected function checkPaymentStatusAndUpdateOrder($charge, $order) {
      * If payment is awaiting to capture then updating order status as 'payment_review'. In case of 3-D secured payment,
      * it will be 'processing'.
      * @param Omise_Gateway_Model_Order $order
+     * @param Omise_Gateway_Model_Api_Charge $charge
      * @return Mage_Core_Controller_Varien_Action
      */
-    protected function paymentAwaiting($order) {
+    protected function paymentAwaiting($order, $charge) {
+        if($charge->isUnauthorized() && $order->getStatus() != Mage_Sales_Model_Order::STATE_CANCELED)
```
jacstn: It is often unauthorized. We can check status of payment many times, before we decide to restore cart.
Solution could be for user to decide by himself if he want to restore cart
Displaying `Reorder` button could be solution for that.

src/app/code/community/Omise/Gateway/Controller/Base.php
```diff
@@ -109,9 +109,12 @@ protected function checkPaymentStatusAndUpdateOrder($charge, $order) {
...
+            return $this->cancelOrder($order, $charge);
```
guzzilar, line 117: Charge unauthorised does not always mean "cancel". It can be a case of "waiting for the bank issuer to process the payment." Cancelling an order right away sounds incorrect behaviour.

PR #249, opened 2020-04-09
Fixing invoice status issue after CC payment.

mayurkathale:
#### 1. Objective

This PR addresses the issue in payment using credit card that invoice and order status doesn't get updated to be "paid".

This issue occurs whenever 3DS is enabled in Magento and is not enabled on Omise account. So, it doesn’t redirect to any page to enter OTP.
So whenever we have “auth and capture” after payment order invoice status is still ‘pending’, but it should be ‘paid’ instead.

This PR to fix invoice status to be paid after payment.

#### 2. Description of change

Updated model class for CC payment to update status of invoice to be paid.

#### 3. Quality assurance

Omise plugin version: Omise-Magento 1.21
PHP version: 5.6.

#### 4. Impact of the change

Credit card payment should work normally.

#### 5. Priority of change

High

#### 6. Additional Notes

NA

### DIFF
--- a/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
@@ -77,6 +77,7 @@ class Omise_Gateway_Model_Payment_Creditcard extends Omise_Gateway_Model_Payment
                 $charge = $this->processPayment($payment, $invoice->getBaseGrandTotal());
                 if($charge->isSuccessful()) {
                     $invoice->setState(Mage_Sales_Model_Order_Invoice::STATE_PAID)->save();
+                    $state_object->setStatus($order->getConfig()->getStateDefaultStatus(Mage_Sales_Model_Order::STATE_PROCESSING));
                 }
                 $payment->setCreatedInvoice($invoice)
                     ->setIsTransactionClosed(false)
--- a/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
+++ b/src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
@@ -74,12 +74,13 @@ class Omise_Gateway_Model_Payment_Creditcard extends Omise_Gateway_Model_Payment
 
             case Mage_Payment_Model_Method_Abstract::ACTION_AUTHORIZE_CAPTURE:
                 $invoice = $order->prepareInvoice()->register();
-                $invoice->setState(Mage_Sales_Model_Order_Invoice::STATE_PAID)->save();
                 $charge = $this->processPayment($payment, $invoice->getBaseGrandTotal());
-
+                if($charge->isSuccessful()) {
+                    $invoice->setState(Mage_Sales_Model_Order_Invoice::STATE_PAID)->save();
+                }
                 $payment->setCreatedInvoice($invoice)
-                    ->setIsTransactionClosed(true)
-                    ->setIsTransactionPending(false)
+                    ->setIsTransactionClosed(false)
+                    ->setIsTransactionPending(true)
                     ->addTransaction(
                         Mage_Sales_Model_Order_Payment_Transaction::TYPE_CAPTURE,
                         $invoice,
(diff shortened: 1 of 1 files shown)


src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
```diff
@@ -74,12 +74,12 @@ public function initialize($payment_action, $state_object)
 
             case Mage_Payment_Model_Method_Abstract::ACTION_AUTHORIZE_CAPTURE:
                 $invoice = $order->prepareInvoice()->register();
-
+                $invoice->setState(Mage_Sales_Model_Order_Invoice::STATE_PAID)->save();
```
guzzilar, line 77: Just wonder why we set the invoice's state to `paid` before the actual charge has been created?

src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
```diff
@@ -74,12 +74,12 @@ public function initialize($payment_action, $state_object)
...
                 $charge = $this->processPayment($payment, $invoice->getBaseGrandTotal());
 
                 $payment->setCreatedInvoice($invoice)
-                    ->setIsTransactionClosed(false)
-                    ->setIsTransactionPending(true)
+                    ->setIsTransactionClosed(true)
+                    ->setIsTransactionPending(false)
```
guzzilar, line 82: It was a super long time since this has been implemented but if I can recall it correctly, the reason that 
`setIsTransactionClosed = false` and `setIsTransactionPending = true` is because 3-D Secure requires buyer to be redirected out to complete their OTP process. On the other hand, at this step, the payment process hasn't completed yet so the transaction stays pending.

(correct me if I misunderstood something, long time no touch...)

guzzilar:
Would it be possible to have more detail on the PR description about the issue that this PR is trying to fix? I'm just a bit lost because this PR trying to solve the invoice status issue but seems it is modifying code in the part of 3-D Secure payment. Not sure how it is related.

src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
mayurkathale: @guzzilar Look like it should be the way it was before.

src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
mayurkathale: @guzzilar  Look like it should be the way it was before.

src/app/code/community/Omise/Gateway/Model/Payment/Creditcard.php
jonrandy: Doesn't seem right. If something goes wrong during charge creation then this will create issues?

guzzilar:
> Disclaimer: I haven't tested this PR, also I kinda already forgot how Magento 1 invoice system work

I'm not sure but just afraid that only patching one spot wouldn't be enough.
As for the 3-D Secure behaviour in its normal flow will go through the plugin's callback url and execute the [`Omise_Gateway_Model_Order::markAsPaid()`](https://github.com/omise/omise-magento/blob/1-stable/src/app/code/community/Omise/Gateway/Model/Order.php#L63) code to manipulate Invoice and Transaction objects' data and its state-status.

Maybe for this issue, the real root cause that we should be fixing is that, implementing a code to bypass all the redirection flow and execute its normal non-3-D Secure flow when the charge transaction is not in `isAwaiting()` status.

As also the code in [`Omise_Gateway_Model_Payment::_process()`](https://github.com/omise/omise-magento/blob/f78b9b78b2ef3c33c8138ace865b3572b1e3c535/src/app/code/community/Omise/Gateway/Model/Payment.php#L67-L69) is now having a condition to not `_setRedirectFlow()` if its charge status is not in `isAwaiting()` condition, meaning that the `auth_only` action will be effected as well (not go through the redirection and not update its order state and status to `processing`).

mayurkathale:
@guzzilar W.r.t your previous comment. I agree that we need to fix for both 3ds and non-3ds type of transactions. I have tested this for non-3ds and unable to test for 3ds using my local setup. As far as I know this issue happens for only non-3ds transactions so I think if we fix at once place  should work.

guzzilar:
@mayurkathale I mean, with typical non-3-D Secure payment, there are lots more that Magento does the magic behind (under `place` and `capture` methods), but because the Omise 3-D secure option is enabled at the store, so Omise plugin bypasses those magics that Magento provided and do our own `initialize()`. So I think to bypass the manual `Omise_Gateway_Model_Payment_Creditcard::initialize()`, we should either 
- Make the non-3-D Secure charge transaction with 3-D Secure enabled store go through the normal redirection (so it will eventually trigger the `Omise_Gateway_Model_Order::markAsPaid()` method)
- Or, bypasses the redirection flow and execute the non-3-D Secure Magento's magic as normal.

guzzilar:
By magic, I mean this part of the code from `Mage_Sales_Model_Order_Payment::capture()`

Screen Shot 2563-04-20 at 09 37 14

This `capture` method will be triggered only if it goes the normal flow (without manual `initialize()`) . But because the 3-D Secure option on a store is enabled, so it will skip all this and execute our `Omise_Gateway_Model_Payment_Creditcard::initialize()` instead.

So I'm just afraid that only fixing one place might not be enough (we may miss some Magento invoice or transaction condition).

(At least, I think we should execute `$invoice->pay()` instead of manually updating its state and status.)

guzzilar:
Although this PR may solve the invoice issue, but I still think the root case is from another place, at the time we introduced `_setRedirectFlow()` method that causing the non-3-D Secure transaction with 3-D Secure enabled-at-store cannot be redirected. 
As we added a condition to check if the charge is in `pending` status.
```php
if ($charge->isAwaitPayment() && !isset($this->_allowRedirect)) {
    $this->_setRedirectFlow($payment, $charge);
}
```

That means the `auth_only` payment will be affected from that change as well.
So, I'm not sure if patching here is a good place. (we're solving a problem that causing by the previous change. Maybe it would be better if we go back to fix at its root cause).

@jonrandy 🆘 need your opinion.

jonrandy:
TBH I haven't really been following what the problem actually is. Would it be resolved by simply doing the redirect on every CC payment? I've always thought this was way simpler

mayurkathale:
@guzzilar @jonrandy Can you please help me to know preferred final solution. I am still unclear about more changes to be done in this PR.

PR #250, opened 2020-04-15
Fixing wrong calculation of installment amount per month showing 

mayurkathale:
#### 1. Objective

On a checkout page, installment amount to be paid every month is calculated wrong as order tax amount isn't included to calculate installment amount.

For example:
_Expected behaviour:_
**Cart Total:** 4200
**Product price:** 4000
**Tax:** 200
**Installment Fees:** 8%
**Duration in month:** 3 months

**Installment per month** = (4200 * 0.008) = 33.6 

_Actual Behaviour:_
**Cart Total:** 4200
**Product price:** 4000
**Tax:** 200
**Installment Fees:** 8%
**Duration in month:** 3 months

**Installment per month** = (4000 * 0.008) = 32

This PR fixes the issue by including order tax amount as grand total to calculate installment fees. 

#### 2. Description of change

Change in javascript file while is responsible to calculate and display installments per month amount.

Screen Shot 2020-04-15 at 13 01 20

#### 3. Quality assurance

**🔧 Environments:**

- **PHP version**: 7.0.16.

**✏️ Details:**

#### 4. Impact of the change

All payment methods should work normally in Magento 2.

#### 5. Priority of change

Normal

#### 6. Additional Notes
NA

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.grand_total;
+                return + window.checkoutConfig.totalsData.base_grand_total;
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return parseInt(window.checkoutConfig.quoteData.grand_total);
+                return + parseInt(window.checkoutConfig.quoteData.grand_total);
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + parseInt(window.checkoutConfig.quoteData.grand_total);
+                return + window.checkoutConfig.quoteData.grand_total;
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.quoteData.grand_total;
+                return parseInt(window.checkoutConfig.quoteData.grand_total);
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.base_grand_total;
+                return + window.checkoutConfig.quoteData.grand_total;
             },
 
             /**
--- a/view/frontend/web/js/view/payment/method-renderer/omise-offline-conveniencestore-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-offline-conveniencestore-method.js
@@ -97,7 +97,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.grand_total;
+                return + window.checkoutConfig.quoteData.grand_total;
             },
 
             /**


guzzilar:
✋  Just wonder, does this miscalculated number occur only at a display on the checkout page, or it also happens to the charge creation step as well? (I just checked in our `PaymentDataBuilder` class and seems it is using an amount from `$order->getGrandTotalAmount()`)

keeratita:
AFAIK, our merchant said it might make the customer confusing. So, my assumption, it's only a display issue.

guzzilar:
👍  @keeratita 
just fyi, we've just found this ticket on Magento 2 repository 
https://github.com/magento/magento2/issues/22925

Now kinda suspect that it might be able to fix by tax-setting itself. (I think @mayurkathale is now working on it)

keeratita:
> 👍 @keeratita
> just fyi, we've just found this ticket on Magento 2 repository
> [magento/magento2#22925](https://github.com/magento/magento2/issues/22925)
> 
> Now kinda suspect that it might be able to fix by tax-setting itself. (I think @mayurkathale is now working on it)

Oh..thanks a lot for your information.. It seems a bug on Magento itself.

keeratita:
our merchant is using Magento 2.3.3 krub.

mayurkathale:
@keeratita This issue looks like from Tax setting at Magento backend. Still need to find setting which can fix this issue.

jsdupuis:
Hi. I'm the merchant who originally reported the problem to Omise. Feel free to ask me any (not too technical) questions. I don't believe it's a Magento setting problem. All our prices are displayed tax inclusive. Only the installment drop down shows prices without tax which is misleading to customers. We have disabled that payment method for the time being to avoid complaints.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.grand_total;
+                return parseInt(window.checkoutConfig.quoteData.grand_total);
```
jonrandy, line 142: A few things:

a) Is `totalsData.grand_total` a total in the correct currency?
b) Is it definitely an integer? Could `parseInt` break stuff here
c) The original code was already ensuring it was ensuring a number was returned - the unary plus does this conversion (coercion) for you

mayurkathale: a) Yes, This returns correct currency.
b) It is an Integer, but stored as string ("300.0000")
c) will update.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.grand_total;
+                return + parseInt(window.checkoutConfig.quoteData.grand_total);
```
jonrandy, line 142: The `parseInt` is unneccesary `return + window.checkoutConfig.totalsData.grand_total;` already does the numeric conversion
image

mayurkathale: @jonrandy Thanks, updated.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -139,7 +139,7 @@ define(
              * @return {integer}
              */
             getTotal: function () {
-                return + window.checkoutConfig.totalsData.grand_total;
+                return + window.checkoutConfig.quoteData.grand_total;
```
guzzilar, line 142: ✋ just wonder what's the difference between `totalsData` and `quoteData`?
Should we also replace all the other places that are using `totalsData` as well?

mayurkathale: In `totalsData`, _base_grand_total_ and _grand_total_ have different values ideally it should be the same for single currency setup. While on the other hand, `quoteData` has values related to quote such as order prices, customer info. It seems like `quoteData` having correct _grand_total_ value so changing it from `totalsData`.

It will be a good idea to change in other parts as well.

jonrandy: So, you're saying `totalsData` has incorrect values? If that's so, what use is it?

guzzilar: ✋  I just dug into Magento Checkout module (under `vendor/magento/` directory) and found some interesting code.

`module-tax/view/frontend/web/template/checkout/summary/grand-total.html` 👇 
```html
<tr class="grand totals incl">
    <th class="mark" scope="row">
        <strong data-bind="i18n: inclTaxLabel"></strong>
    </th>
    <td data-bind="attr: {'data-th': inclTaxLabel}" class="amount">
        <strong><span class="price" data-bind="text: getValue()"></span></strong>
    </td>
</tr>
```
`grand-total.html` displays its price value using `getValue()` function from `module-tax/view/frontend/web/js/view/checkout/summary/grand-total.js` 👇 

```js
define([
    'Magento_Checkout/js/view/summary/abstract-total',
    'Magento_Checkout/js/model/quote',
    'Magento_Catalog/js/price-utils',
    'Magento_Checkout/js/model/totals'
], function (Component, quote, priceUtils, totals) {
    'use strict';

    return Component.extend({
        defaults: {
            isFullTaxSummaryDisplayed: window.checkoutConfig.isFullTaxSummaryDisplayed || false,
            template: 'Magento_Tax/checkout/summary/grand-total'
        },
        totals: quote.getTotals(),

        ...

        /**
         * @return {*|String}
         */
        getValue: function () {
            var price = 0;

            if (this.totals()) {
                price = totals.getSegment('grand_total').value;
            }

            return this.getFormattedPrice(price);
        },
```
The `totals` variable coming from `Magento_Checkout/js/model/quote` 👇 
```js
define([
    'ko',
    'underscore',
    'domReady!'
], function (ko, _) {
    'use strict';

    /**
     * Get totals data from the extension attributes.
     * @param {*} data
     * @returns {*}
     */
    var proceedTotalsData = function (data) {
            if (_.isObject(data) && _.isObject(data['extension_attributes'])) {
                _.each(data['extension_attributes'], function (element, index) {
                    data[index] = element;
                });
            }

            return data;
        },
        billingAddress = ko.observable(null),
        shippingAddress = ko.observable(null),
        shippingMethod = ko.observable(null),
        paymentMethod = ko.observable(null),
        quoteData = window.checkoutConfig.quoteData,
        basePriceFormat = window.checkoutConfig.basePriceFormat,
        priceFormat = window.checkoutConfig.priceFormat,
        storeCode = window.checkoutConfig.storeCode,
        totalsData = proceedTotalsData(window.checkoutConfig.totalsData),
        totals = ko.observable(totalsData),
        collectedTotals = ko.observable({});

    return {
        totals: totals,
        ...
}
```
Which, referring to `window.checkoutConfig.totalsData`:
```js
totalsData = proceedTotalsData(window.checkoutConfig.totalsData),
totals = ko.observable(totalsData),
```

---

Also, from the code at `vendor/magento/module-checkout/Model/DefaultConfigProvider.php` 👇 

```php
/**
 * Retrieve quote data
 *
 * @return array
 */
private function getQuoteData()
{
    $quoteData = [];
    if ($this->checkoutSession->getQuote()->getId()) {
        $quote = $this->quoteRepository->get($this->checkoutSession->getQuote()->getId());
        $quoteData = $quote->toArray();
        $quoteData['is_virtual'] = $quote->getIsVirtual();

        if (!$quote->getCustomer()->getId()) {
            /** @var $quoteIdMask \Magento\Quote\Model\QuoteIdMask */
            $quoteIdMask = $this->quoteIdMaskFactory->create();
            $quoteData['entity_id'] = $quoteIdMask->load(
                $this->checkoutSession->getQuote()->getId(),
                'quote_id'
            )->getMaskedId();
        }
    }
    return $quoteData;
}
```

and 
```php
private function getTotalsData()
{
    /** @var \Magento\Quote\Api\Data\TotalsInterface $totals */
    $totals = $this->cartTotalRepository->get($this->checkoutSession->getQuote()->getId());
    $items = [];
    /** @var  \Magento\Quote\Model\Cart\Totals\Item $item */
    foreach ($totals->getItems() as $item) {
        $items[] = $item->__toArray();
    }
    $totalSegmentsData = [];
    /** @var \Magento\Quote\Model\Cart\TotalSegment $totalSegment */
    foreach ($totals->getTotalSegments() as $totalSegment) {
        $totalSegmentArray = $totalSegment->toArray();
        if (is_object($totalSegment->getExtensionAttributes())) {
            $totalSegmentArray['extension_attributes'] = $totalSegment->getExtensionAttributes()->__toArray();
        }
        $totalSegmentsData[] = $totalSegmentArray;
    }
    $totals->setItems($items);
    $totals->setTotalSegments($totalSegmentsData);
    $totalsArray = $totals->toArray();
    if (is_object($totals->getExtensionAttributes())) {
        $totalsArray['extension_attributes'] = $totals->getExtensionAttributes()->__toArray();
    }
    return $totalsArray;
}
```
If we dig down to `cartTotalRepository`, it seems both are  initiating the Quote repository the same `vendor/magento/module-quote/Model/GuestCart/GuestCartRepository.php` 👇 
```php
public function get($cartId)
{
    /** @var $quoteIdMask QuoteIdMask */
    $quoteIdMask = $this->quoteIdMaskFactory->create()->load($cartId, 'masked_id');
    return $this->quoteRepository->get($quoteIdMask->getQuoteId());
}
```

---
I'm not sure where it will lead us to... but if the origin of these 2 variable (window.checkoutConfig.quoteData) and (window.checkoutConfig.totalsData) are coming from Quote object, then probably it should return the same result? 🤔 

I was also searching and found no place in Magento project that access to `window.checkoutConfig` directly too. Maybe using data from `window.checkoutConfig` directly might cause this issue? (not sure...)

What if we check what is inside `quote.totals()` or, simply using the price from `Magento_Checkout/js/model/quote` instead of accessing directly to the global `window.checkoutConfig` variable?

guzzilar: Sorry for the long comment 🙇

guzzilar: Btw, also at `vendor/magento/module-tax/view/frontend/web/js/view/checkout/summary/grand-total.js`
Not sure how it works but looking on these 2 functions and it really seems that `totalsData.grand_total` does not include tax.
```js
/**
 * @return {*|String}
 */
getValue: function () {
    var price = 0;

    if (this.totals()) {
        price = totals.getSegment('grand_total').value;
    }

    return this.getFormattedPrice(price);
},

/**
 * @return {*}
 */
getGrandTotalExclTax: function () {
    var total = this.totals();

    if (!total) {
        return 0;
    }

    return this.getFormattedPrice(total['grand_total']);
},
```

mayurkathale: @jonrandy I would say not expected values more than incorrect values as still don't know the reason why they have excluded tax amount in grand price.

mayurkathale:
@jonrandy anything would you like to add to this PR?

PR #261, opened 2020-06-11
Adding billing address information to card while creating token

mayurkathale:
#### 1. Objective

This PR adds billing address information to charge while creating token. If customer pays using card payment, billing address information should be stored in charge.

Checkout Page:
Screen Shot 2020-06-11 at 12 07 59

Omise Dashboard:
Screen Shot 2020-06-11 at 12 37 31

**Related information**:
Related issue(s): # https://phabricator.omise.co/T13092

#### 2. Description of change

- Updated Javascript file to fetch information from quote billing address and store in `card` object.

#### 3. Quality assurance

**🔧 Environments:**
- **Platform version**: Magento CE 2.3.4.
- **Omise plugin version**: Omise-Magento 2.11.
- **PHP version**: 7.2.1.

**✏️ Details:**

- Add product to card
- go to checkout page.
- place order using card payment.
- After order placement check charge created in omise dashboard, it should have appropriate billing addres information

#### 4. Impact of the change
All payment methods should work normally.

#### 5. Priority of change

Normal

#### 6. Additional Notes

NA

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
@@ -263,16 +263,14 @@ define(
                 var selectedBillingAddress = quote.billingAddress();
                 var address = {
                     state          : selectedBillingAddress.region,
-                    postal_code: selectedBillingAddress.postcode,
-                    phone_number: selectedBillingAddress.telephone,
-                    country: selectedBillingAddress.countryId,
-                    city: selectedBillingAddress.city
+                    postal_code    : selectedBillingAddress.postcode,
+                    phone_number   : selectedBillingAddress.telephone,
+                    country        : selectedBillingAddress.countryId,
+                    city           : selectedBillingAddress.city
                 }
-                if(selectedBillingAddress.street.length) {
-                    address.street1 = selectedBillingAddress.street[0]
-                    if(selectedBillingAddress.street[1]) {
-                        address.street2 = selectedBillingAddress.street[1]
-                    }
+                address.street1 = selectedBillingAddress.street[0]
+                if(selectedBillingAddress.street[1]) {
+                    address.street2 = selectedBillingAddress.street[1]
                 }
                 return address
             }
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
```diff
@@ -257,6 +257,24 @@ define(
                             redirectOnSuccessAction.execute();
                         }
                     });
+            },
+
+            getSelectedTokenBillingAddress: function() {
+                var selectedBillingAddress = quote.billingAddress();
+                var address = {
+                    state          : selectedBillingAddress.region,
```
danfowler, line 265: Do we have existing formatting rules for this?

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
```diff
@@ -257,6 +257,24 @@ define(
...
+                    postal_code: selectedBillingAddress.postcode,
+                    phone_number: selectedBillingAddress.telephone,
+                    country: selectedBillingAddress.countryId,
+                    city: selectedBillingAddress.city
+                }
+                if(selectedBillingAddress.street.length) {
+                    address.street1 = selectedBillingAddress.street[0]
```
danfowler, line 272: Do we need to assign these `street*` attributes conditionally?  Would it be bad if we just assigned them as-is in the `address` var above since JavaScript has no issue with "out of bounds" 🤷‍♂️?

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
danfowler: Thanks

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
mayurkathale: `selectedBillingAddress.street` here it can be ["street1"] or ["street1", "street2"] which depends on how user fill in billing address. I'll remove first "if" as street address is mandatory but keep second condition to check if it exists.

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
mayurkathale: Done. Reformatting this as it should be more readable.

view/frontend/web/js/view/payment/method-renderer/omise-cc-method.js
danfowler: 🙏

PR #316, opened 2021-11-29
[FRON-987] - Use Expired Bool from Api to check charge expiry

timothyOmise:
Use 'expired' bool value returned from api to check is charges need to be cancelled during cron sync

#### 1. Objective

Fix issue where Magento users that did not use UTC time as their default time were experiencing issue with charges being expired when they should not have been.

**Related information**:

https://omise.atlassian.net/browse/FRON-987

#### 2. Description of change

Added new method to check if charge is expired. The 'expired' value from the Api is now used instead of checking expired time against local server now time

#### 3. Quality assurance

Test charges are expired correctly

**🔧 Environments:**

 **Platform version**: Magento CE 2.1.5.
**Omise plugin version**: Omise-Magento 2.1.
 **PHP version**: 7.0.16.

**✏️ Details:**

#### 4. Impact of the change

None

#### 5. Priority of change

High

### DIFF
--- a/Cron/OrderSyncStatus.php
+++ b/Cron/OrderSyncStatus.php
@@ -159,9 +159,8 @@ class OrderSyncStatus
             foreach ($orderIds as $order) {
                 $this->lastProcessedOrderId = $order['entity_id'];
                 $this->order = $this->orderRepository->get($order['entity_id']);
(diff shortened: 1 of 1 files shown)


kan98 (approved):
tested on my machine and it can still expire charges correctly 👍🏽

PR #326, opened 2022-02-18
[FRON-972]

timothyOmise:
#### 1. Objective

Currently the magento plugin has 2 flows that will create and send invoices when merchant is using 3ds (redirect & webook). This PR aims to mitigate this issue by delaying the creation and sending of the invoices by 10secs when initiated via webhook. This delay should allow time for any redirect process to complete and update the order records to reflect that the invoice for the order has been created and sent already.

#### 2. Description of change

A delay has been added above the 'sendInvoiceEmail' method that is called in the PendingInvoiceHandler Class handler method. The creation of the invoice from the PendingInvoiceHandler Class handler method has been moved inside the 'sendInvoiceEmail' method. A 2nd argument has been added to the 'sendInvoiceEmail' to flagh whether or not to created the invoice before sending. 

A slight updated to 'Omise\Payment\Model\Api\Capabilities' class has been made. The purpose of this is to pass the exception message to the custom exception 'LocalizedException' to allow for better error debugging for interrogations. 

#### 3. Quality assurance

3DS checkout process needs to be tested. Ensure that only 1 invoice is created and sent. As we fixing a race condition there may still be very rare edge cases that still send and create 2 invoices. Therefore a 99% success rate is acceptable.

**🔧 Environments:**

Local php development env

i.e.
- **Platform version**: Magento CE 2.4.3.
- **Omise plugin version**: Omise-Magento 2.20.5.
- **PHP version**: 7.4.27.

**✏️ Details:**

Created a order with 3DS enabled and conffirm that only 1 invoice is created and sent

#### 4. Impact of the change

N/A

#### 5. Priority of change

High

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # CHANGELOG
 
+## [v2.20.5 _(Feb 18, 2022)_](https://github.com/omise/omise-magento/releases/tag/v2.20.5)
+
+### 👾 Bug Fixes
+- Delay creation and sending of invoice when initiated via webook (PR [#326](https://github.com/omise/omise-magento/pull/326))
+
 ## [v2.20.4 _(Feb 14, 2022)_](https://github.com/omise/omise-magento/releases/tag/v2.20.4)
 
 ### 👾 Bug Fixes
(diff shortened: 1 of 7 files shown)


aashishgurung:
Already fixed this in other branch.

PR #337, opened 2022-04-18
[APM2-578] Fetch capabilities

new4762:
#### 1. Objective

Currently the payment options that display in checkout page didn’t fetch from capabilities api 
so this PR focus to update checkout page to display payment options correctly 
for `installment` , `internet banking`  and `mobile banking`

- Update installment to display from fetching capabilities api  
- Update internet banking to display from fetching capabilities api  
- Update mobile banking to display from fetching capabilities api  
- Update mobile banking to use currency options from capabilities api  

This section will be used in the release notes. 

**Related information**:
Related issue(s): #< GitHub ticket number > (optional)

- https://omise.atlassian.net/browse/APM2-690
- https://omise.atlassian.net/browse/APM2-691
- https://omise.atlassian.net/browse/APM2-692

#### 2. Description of change

Replace statics display payments options to be dynamic display from capabilities api .
effect to `installment` , `internet banking`  and `mobile banking`

 
#### 3. Quality assurance

Test payments option displays correctly and  can created charge to staging 

**🔧 Environments:**

- **Platform version**: Magento CE 2.2.3
- **Omise plugin version**: Omise-Magento 2.23.1
- **Omise plugin version**: Omise-php v2.13
- **PHP version**: 7.1

**✏️ Details:**

Try switching `source_type` config to check is really dynamic display and testing create charges to make sure it correct sources type
 
#### 4. Impact of the change

`installment` , `internet banking`  and `mobile banking` will display only enabled sources type

**Internet banking**
image
image

**Installment**
image
image
image

**Mobile banking**
image
image

Note: Please provide a screenshot if your changed impact to UI.

#### 5. Priority of change

Normal

#### 6. Additional Notes

Any further information that you would like to add.

PR #339, opened 2022-04-26
[FRON-1432]: Preventing Omise plugin to initialize when public and secret keys are not present

aashishgurung:
#### 1. Objective

Fix the issue on checkout page when keys are not configured. In multi-store environment, if a merchant uses Omise payment gateway for one store only and another payment gateway (like Stripe) for other sub-store, the checkout page throws an error when merchant does not configure Omise keys for the sub-store in which they not using Omise payment gateway.

Jira Ticket:
https://omise.atlassian.net/browse/FRON-1432

#### 2. Description of change

A method `initOmise` is added which gets called in the `construct` method of `/omise-magento/Model/Api/Capabilities.php`. The `initOmise` method will fetch the public and secret keys of the current store. It stops the initialization of the plugin if one of the keys is not configured.

#### 3. Quality assurance

Set up two stores, let say one for Malaysia and another for Singapore. 

**Default store (MY),**
- Credit Debit cards (Stripe)
- FPX (Stripe)

**Sub-store (SG)**
- Paynow (Omise)
- Credit Debit cards (Omise)

After the setup, go to the store where Omise payment is not setup. Try adding an item to the cart and go to the checkout page. Before the changes of this PR, you should see blank page or an error page. Switching back to this change, you should see a checkout page and should be able to continue the process.

**🔧 Environments:**

Specify the details of your test environments, including, for each, the platform version (on which the plugin was run), the Omise plugin version, and the versions of your system software such as PHP or Ruby.

- Platform version: Magento CE 2.4.3
- Omise plugin version: Omise-Magento 2.23.2
- PHP version: 7.1.15

### DIFF
--- a/Model/Api/Capabilities.php
+++ b/Model/Api/Capabilities.php
@@ -33,7 +33,7 @@ class Capabilities extends BaseObject
         $this->config->setStoreId($storeId);
 
         // Initialize only if both keys are present
-        if(!$this->config->getPublicKey() || !$this->config->getSecretKey()) {
+        if (!$this->config->getPublicKey() || !$this->config->getSecretKey()) {
             return false;
         }
 
--- a/Model/Api/Capabilities.php
+++ b/Model/Api/Capabilities.php
@@ -4,13 +4,39 @@ namespace Omise\Payment\Model\Api;
 
 use Magento\Framework\Exception\LocalizedException;
 use OmiseCapabilities;
+use Omise\Payment\Model\Config\Config;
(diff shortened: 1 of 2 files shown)


kan98 (approved):
would be good to start having test coverage in these models I think. Maybe when you have some time 🙂

new4762 (approved):
if you want to release today kindly add my PR -> https://github.com/omise/omise-magento/pull/337 with you release plse 🙏

PR #341, opened 2022-05-05
[APM2-715] Move OCBC Pao as standalone payment

new4762:
#### 1. Objective

Currently OCBC requested us to move OCBC PAO out from under Mobile Banking options to standalone payment
So this PR will make a new payment option `OCBC PAO` for magento and remove it from mobile banking

This section will be used in the release notes. 

**Related information**:
Related issue(s): https://omise.atlassian.net/browse/APM2-715

#### 2. Description of change

- add new offsite payment for `ocbc pao`
- remove `ocbc pao` from mobile banking 
- set mobile banking display only when checkout is THB

#### 3. Quality assurance

OCBC PAO can make a charge with IOS or Android successfully

**🔧 Environments:**

- **Platform version**: Magento CE 2.2.3
- **Omise plugin version**: Omise-Magento 2.23.2
- **Omise plugin version**: Omise-php v2.13
- **PHP version**: 7.1

**✏️ Details:**

- Using PSP SG for testing
- Enable `mobile_banking_ocbc_pao` source type
- Enable OCBC PAO option on magento admin panel
- make a pay with  OCBC PAO
- charge should created successfully 

#### 4. Impact of the change

image

image

image

image

#### 5. Priority of change

Normal

#### 6. Additional Notes

Any further information that you would like to add.

PR #342, opened 2022-05-06
[FRON-1442]: Issue of new order email not being sent when order is placed with non omise payment method has been fixed.

aashishgurung:
#### 1. Objective

Fix the issue of new order email not being sent when order is placed with non omise payment method.

**Jira Ticket**: https://omise.atlassian.net/browse/FRON-1442

#### 2. Description of change

The plugin will set `setCanSendNewEmailFlag` to false only if the payment method is offered by our plugin. Added a new method `isOmisePayment` in the `OmiseHelper` class that will check the passed payment method to an array of payment methods provided by our plugin. If the payment method is non Omise payment method then `isOmisePayment` will return false and the code does nothing for `setCanSendNewEmailFlag` method and vice versa.

#### 3. Quality assurance

Configure email in the Magento. Go to the frontend and checkout with an item. Choose a non Omise payment method like Bank Transfer Payment, Cash on Delivery or Stripe payment methods. You should receive a new order email.

**🔧 Environments:**

- Platform version: Magento CE 2.4.3
- Omise plugin version: Omise-Magento 2.23.2
- PHP version: 7.1.15

### DIFF
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -310,15 +310,15 @@ class OmiseHelper extends AbstractHelper
      */
     public function isOmisePayment($paymentMethod)
     {
-        return in_array(
-            $paymentMethod,
-            array_merge(
-                array_merge($this->offsitePaymentMethods, $this->offsitePaymentImageCode),
-                [
-                    Config::CODE,
-                    Conveniencestore::CODE
-                ]
-            )
+        $offsitePaymentMethods = array_merge($this->offsitePaymentMethods, $this->offsitePaymentImageCode);
+        $omisePaymentMethods = array_merge(
+            $offsitePaymentMethods,
+            [
+                Config::CODE,
+                Conveniencestore::CODE
+            ]
         );
+
+        return in_array($paymentMethod, $omisePaymentMethods);
     }
 }
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -32,6 +32,37 @@ class OmiseHelper extends AbstractHelper
     protected $header;
 
     /**
+     * @var array
+     */
+    private $offsitePaymentMethods = [
+        Alipay::CODE,
+        Internetbanking::CODE,
+        Installment::CODE,
+        Truemoney::CODE,
+        Pointsciti::CODE,
+        Fpx::CODE,
+        Alipayplus::ALIPAY_CODE,
+        Alipayplus::ALIPAYHK_CODE,
+        Alipayplus::DANA_CODE,
+        Alipayplus::GCASH_CODE,
+        Alipayplus::KAKAOPAY_CODE,
+        Alipayplus::TOUCHNGO_CODE,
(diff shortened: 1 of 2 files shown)


new4762 (approved):
Look good but about helper list my https://github.com/omise/omise-magento/pull/341 will add another payment option . so  i think is will need to update that list as well

aashishgurung:
> #341

Thanks for the update. I will make the changes once your PR is merged.

Helper/OmiseHelper.php
```diff
@@ -292,4 +293,40 @@ public function createInvoiceAndMarkAsPaid($order, $chargeId, $isCapture = true)
...
+
+    /**
+     * This method checks and return TRUE if $paymentMethod is
+     * offered by our Omise payment plugin otherwise returns false.
+     *
+     * @param string $paymentMethod
+     *
+     * @return boolean
+     */
+    public function isOmisePayment($paymentMethod)
```
kan98, line 305: would it be better to pull lists from `isPayableByImageCode` and `isOffsitePayment` and then add credit card? so we don't have multiple lists to have to maintain

aashishgurung: It would be better. Thanks.

Helper/OmiseHelper.php
```diff
@@ -292,4 +299,26 @@ public function createInvoiceAndMarkAsPaid($order, $chargeId, $isCapture = true)
...
+    {
+        $offsitePaymentMethods = array_merge($this->offsitePaymentMethods, $this->offsitePaymentImageCode);
+        $omisePaymentMethods = array_merge(
+            $offsitePaymentMethods,
+            [
+                Config::CODE,
+                Conveniencestore::CODE
```
kan98, line 318: I think convenience store could also be a part of offsite payment methods. it must have been missed before 🤔

kan98 (approved):

Helper/OmiseHelper.php
```diff
@@ -30,6 +31,37 @@ class OmiseHelper extends AbstractHelper
...
+        Mobilebanking::CODE,
+        Rabbitlinepay::CODE,
+    ];
+
+    /**
+     * Payment method payable by image code
+     *
+     * @var array
+     */
+    private $offsitePaymentImageCode = [
```
hpariwat, line 59: Probably just terminology but normally payment methods with image codes are referred to as offlines rather than offsites. Or I guess we could rename it to just `paymentImageCode` to avoid confusion.

aashishgurung: Let's go with `offlinePaymentMethods` and drop the `ImageCode` if it makes sense. This will be consistent with the other variable name `offsitePaymentMethods.

hpariwat: One thing to note though, there are also offline payment methods without image codes (like `Conveniencestore::CODE`), so may need a comment to clarify that this is only for offline payments with image code feature.

PR #344, opened 2022-05-12
[EM-1799] Add GrabPay payment method

kan98:
#### 1. Objective

Explain in non-technical terms **WHY this PR is required**.
E.g.: What feature it adds, what problem it solves...

Added a new payment method, GrabPay, for SG and MY merchants.

Screen Shot 2565-05-12 at 15 14 48

Screen Shot 2565-05-12 at 15 15 47

This section will be used in the release notes. 

**Related information**:
Related issue(s): #< GitHub ticket number > (optional)

#### 2. Description of change

A general description of **WHAT changed in the codebase**, but short of an English version of the diff. Assume that people reading this will also be looking at the output of `git diff` and guide them to the highlights.

Additionally add the reasoning for change details if they're complex or abstract.

#### 3. Quality assurance

tested on my development environment

**✏️ Details:**

Explain how to manually test this feature.
For example if changes were made in the UI or in the API, explain where and if any specific access is needed.

#### 5. Priority of change

Normal, High or Immediate.

Normal

#### 6. Additional Notes

Any further information that you would like to add.

### DIFF
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -19,6 +19,7 @@ use Omise\Payment\Model\Config\Alipayplus;
 use Omise\Payment\Model\Config\Mobilebanking;
 use Omise\Payment\Model\Config\Rabbitlinepay;
 use Omise\Payment\Model\Config\Ocbcpao;
+use Omise\Payment\Model\Config\Grabpay;
 use Magento\Framework\Exception\LocalizedException;
 use Omise\Payment\Helper\OmiseHelper;
 use Omise\Payment\Helper\OmiseEmailHelper;
@@ -213,6 +214,9 @@ class Offsite extends Action
                         break;
                     case Ocbcpao::CODE:
                         $dispPaymentMethod = "OCBC Pay Anyone";
+                        break;
+                    case Grabpay::CODE:
+                        $dispPaymentMethod = "GrabPay";
                         break;
                 }
                 
--- a/Gateway/Request/APMBuilder.php
+++ b/Gateway/Request/APMBuilder.php
@@ -19,6 +19,7 @@ use Omise\Payment\Model\Config\Alipayplus;
 use Omise\Payment\Model\Config\Mobilebanking;
 use Omise\Payment\Model\Config\Rabbitlinepay;
 use Omise\Payment\Model\Config\Ocbcpao;
+use Omise\Payment\Model\Config\Grabpay;
 
 use Omise\Payment\Observer\ConveniencestoreDataAssignObserver;
 use Omise\Payment\Observer\FpxDataAssignObserver;
@@ -230,6 +231,12 @@ class APMBuilder implements BuilderInterface
             case Ocbcpao::CODE:
                 $paymentInfo[self::SOURCE] = [
                     self::SOURCE_TYPE => 'mobile_banking_ocbc_pao',
+                    self::PLATFORM_TYPE => $this->helper->getPlatformType(),
+                ];
+                break;
+            case Grabpay::CODE:
+                $paymentInfo[self::SOURCE] = [
+                    self::SOURCE_TYPE => 'grabpay',
                     self::PLATFORM_TYPE => $this->helper->getPlatformType(),
                 ];
                 break;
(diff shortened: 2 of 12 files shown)


Gateway/Request/APMBuilder.php
```diff
@@ -233,6 +234,12 @@ public function build(array $buildSubject)
                     self::PLATFORM_TYPE => $this->helper->getPlatformType(),
                 ];
                 break;
+            case Grabpay::CODE:
+                $paymentInfo[self::SOURCE] = [
+                    self::SOURCE_TYPE => 'grabpay',
+                    self::PLATFORM_TYPE => $this->helper->getPlatformType(),
```
hpariwat, line 240: Does it use a platform type? 🤔

kan98: I figured there's no harm in sending it (maybe the rms implementation might need it). and it looks like most other APMs are sending this field through too

new4762: Can confirm about rms -> yep will need to send `PLATFORM_TYPE` to RMS channel

hpariwat (approved):

view/frontend/web/template/payment/offsite-grabpay-form.html
```diff
@@ -0,0 +1,54 @@
...
+            data-bind="attr: {
+               'id': getCode()
+               },
+               value: getCode(),
+               checked: isChecked,
+               click: selectPaymentMethod,
+               visible: isRadioButtonVisible(),
+               enable: isActive()" />
+        <label data-bind="attr: {'for': getCode()}" class="label">
+            <span data-bind="text: getTitle()"></span>
```
new4762, line 14: Do you mind ask PO about adding the logo 🤔 , i think we should start adding the logo for both magento and wc

new4762 (approved):

view/frontend/web/template/payment/offsite-grabpay-form.html
kan98: Why do you think we should start adding logos? If we do, then we should go back and add logos for all previous methods too, I think, to be consistent. But there doesn't seem to be a product requirement for it 🤔

tookaa2 (approved):

view/frontend/web/template/payment/offsite-grabpay-form.html
new4762: Is bc i seen Android and IOS will alway have the logo right ? so imo we should follow that as well

Controller/Callback/Offsite.php
```diff
@@ -214,6 +215,9 @@ public function execute()
                     case Ocbcpao::CODE:
                         $dispPaymentMethod = "OCBC Pay Anyone";
                         break;
+                    case Grabpay::CODE:
+                        $dispPaymentMethod = "GrabPay";
```
aashishgurung, line 219: I see if we add a payment method then we need to add a display method as well. For future improvements, we can extract display payment method text to individual payment class under config and have a factory class that will return a class instance according to the payment code. Then we can simply call `$class->getDisplayPaymentMethod()`.

aashishgurung (approved):

FhanOmise (approved):

Controller/Callback/Offsite.php
kan98: yes I agree, that's a good idea. Maybe we can have a PR to refactor this, including the older payment methods as well. Will keep that in mind 👍🏽

view/frontend/web/template/payment/offsite-grabpay-form.html
kan98: I think there's a bit more to it than that. In this case we follow magento's standards as merchants may use our plugins alongside other payment providers. So in theory, they should be consistent with each other. But it's not a bad idea, maybe it'll be a good point to raise with product 🙂

new4762: Sure, NP 👍

som-m (approved):
👍

image

image

image

PR #357, opened 2022-06-14
Feature/fron 1521

aashishgurung:
#### 1. Objective

Prevent cancelation of order when a user enters the return URI directly in the browser tab in the middle of the transaction.

This section will be used in the release notes. 

Jira Ticket: [#1521](https://opn-ooo.atlassian.net/browse/FRON-1521)

#### 2. Description of change

Added a logic that will check the `REFERER` header to determine whether the `return_uri` was hit by the Omise server or directly by the user. If the `return_uri` is hit by the user then s/he will be directed to the cart page without affecting the order status.

#### 3. Quality assurance

- Setup to use manual capture
- Add items to cart
- Checkout via Card. You will be redirected to the authorize page
- Do nothing on authorize page. Open a new tab
   - Go to the return URI `http://{YOUR_BASE_PATH}/omise/callback/threedsecure`
   - You should be redirected to the _**cart page**_.
 - Go to the authorize page and complete the transaction
 - You should be redirected to the t_**hank you page**_.

After completing these steps, check the **_order status_**. It should not be **_cancelled_**.

_**Cart page**_ (Should be displayed if a user directly access the return URI)
Screen Shot 2565-06-14 at 12 06 58

_**Thank you page**_ (Should be displayed after the payment is complete)
Screen Shot 2565-06-14 at 12 37 16

**_Order page_**
Screen Shot 2565-06-14 at 12 40 37

**🔧 Environments:**

Omise plugin version: Omise-Magento 2.26.0
PHP version: 7.4.28

### DIFF
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -366,13 +366,13 @@ class OmiseHelper extends AbstractHelper
     }
 
     /**
-     * Validate whether the a URI was triggered by Omiser server or not
+     * Validate whether the a URI was triggered by Omise server or not
      */
     public function validate3DSReferer()
     {
         $refererValue = $this->header->getHttpReferer();
         $isProduction = strpos($refererValue, 'https://api.omise.co') === 0;
-        $isStaging = strpos($this->header->getHttpReferer(), 'https://api.staging-omise.co') === 0;
+        $isStaging = strpos($refererValue, 'https://api.staging-omise.co') === 0;
 
         return $isProduction || $isStaging;
     }
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -364,4 +364,16 @@ class OmiseHelper extends AbstractHelper
     {
         return in_array($paymentMethod, $this->cardPaymentMethods);
     }
+
+    /**
+     * Validate whether the a URI was triggered by Omiser server or not
+     */
+    public function validate3DSReferer()
+    {
+        $refererValue = $this->header->getHttpReferer();
+        $isProduction = strpos($refererValue, 'https://api.omise.co') === 0;
+        $isStaging = strpos($this->header->getHttpReferer(), 'https://api.staging-omise.co') === 0;
+
+        return $isProduction || $isStaging;
+    }
 }
--- a/Controller/Callback/Threedsecure.php
+++ b/Controller/Callback/Threedsecure.php
@@ -203,6 +203,14 @@ class Threedsecure extends Action
             return (new CaptureResultValidator)->validate($charge);
         }
 
+        if ($this->config->isWebhookEnabled()) {
+            // preventing the order being canceled when the return_uri is hit directly
+            // by the user and not by the Omise server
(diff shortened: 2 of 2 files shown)


Helper/OmiseHelper.php
```diff
@@ -364,4 +364,16 @@ public function isCreditCardPaymentMethod($paymentMethod)
...
     }
+
+    /**
+     * Validate whether the a URI was triggered by Omiser server or not
+     */
+    public function validate3DSReferer()
+    {
+        $refererValue = $this->header->getHttpReferer();
+        $isProduction = strpos($refererValue, 'https://api.omise.co') === 0;
+        $isStaging = strpos($this->header->getHttpReferer(), 'https://api.staging-omise.co') === 0;
```
hpariwat, line 375:
```suggestion
        $isStaging = strpos($refererValue, 'https://api.staging-omise.co') === 0;
```

Helper/OmiseHelper.php
```diff
@@ -364,4 +364,16 @@ public function isCreditCardPaymentMethod($paymentMethod)
     {
         return in_array($paymentMethod, $this->cardPaymentMethods);
     }
+
+    /**
+     * Validate whether the a URI was triggered by Omiser server or not
```
hpariwat, line 369:
```suggestion
     * Validate whether the a URI was triggered by Omise server or not
```

Helper/OmiseHelper.php
aashishgurung: @hpariwat Thanks

kan98 (commented):
should we cover the cases for Offsite APM callbacks as well? they follow a similar flow to 3ds payments.

aashishgurung:
> should we cover the cases for Offsite APM callbacks as well? they follow a similar flow to 3ds payments.

@kan98  If user can mess around like in 3DS flow then absolutely. Let's create a separate ticket for it.

kan98 (approved):

new4762 (approved):

tanawin-opn:
Tested and verified krub

But if the flow incidentally redirect customer to return_uri and can't go back to authorize_uri
does it mean that the customer always have to add the cart all over again? 
because it's look like magento already empty the cart.

aashishgurung:
> Tested and verified krub
> 
> But if the flow incidentally redirect customer to return_uri and can't go back to authorize_uri does it mean that the customer always have to add the cart all over again? because it's look like magento already empty the cart.

@tanawin-opn Our plugin redirects user to the return_uri only after completing or cancelling the transaction from authorize_uri. As you know, on completion it shows thank you page and on cancellation it shows cart page with items.

The situation you describing occurs when a user breaks out of the flow and access return_uri without doing anything on the authorize_uri page. Users are not suppose to do that.

I tried re-adding the items into the cart when user access return_uri but by doing so, it deleted the active order in session and if we try to complete or cancel the order after accessing the return_uri then it will display order does not exist.

So, as per my understanding, our flow will not trigger this is a case. It occurs if user tries to mess with the flow. If you think we need to look into this then let's create a separate ticket for this.

jutinapas (approved):

PR #359, opened 2022-06-29
[FRON-1536]: Refactored the code. Added a logic to reverse the cancel…

aashishgurung:
#### 1. Objective

Update the order status when charge status is updated manually by Operations Team.

Jira Ticket: [#1536](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-247)

#### 2. Description of change

Added a logic to reverse the cancelled order and set the status as processing when the order status is cancelled and the charge is successful. This will be triggered via the webhook. Refactored the `Complete.php` file.

#### 3. Quality assurance

- Checkout with any item
- On reaching the authorize page,
   - First, test by authorizing the payment
   - Second, test by not authorizing the payment. Simply close the page once you reach there.
- On Magento admin, find the order and cancel it manually.
   - You can cancel the order from Omise admin dashboard by marking the charge as failed.
- On Omise admin dashboard, find the charge and mark it as a success.
- On Magento admin, check the order you just cancelled. It status should be set to processing and appropriate comment should be there in the comment history.

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.26.0
- PHP version: 7.4.28

### DIFF
--- a/Plugin/ConfigSectionPaymentPlugin.php
+++ b/Plugin/ConfigSectionPaymentPlugin.php
@@ -69,6 +69,7 @@ class ConfigSectionPaymentPlugin
     {
         $configFields = $configData['groups']['omise']['fields'];
 
+        // if sandbox status is updated the updated value will be under 'value' key else it won't have the value key
         $sandboxStatus = array_key_exists('value', $configFields['sandbox_status'])
             ? $configFields['sandbox_status']['value']
             : $configFields['sandbox_status'];
--- a/Plugin/ConfigSectionPaymentPlugin.php
+++ b/Plugin/ConfigSectionPaymentPlugin.php
@@ -69,7 +69,9 @@ class ConfigSectionPaymentPlugin
     {
         $configFields = $configData['groups']['omise']['fields'];
 
-        $standboxStatus = array_key_exists('value', $configFields['sandbox_status']) ? $configFields['sandbox_status']['value'] : $configFields['sandbox_status'];
+        $standboxStatus = array_key_exists('value', $configFields['sandbox_status'])
+            ? $configFields['sandbox_status']['value']
+            : $configFields['sandbox_status'];
 
         $publicKeyIndex = $standboxStatus ? 'test_public_key' : 'live_public_key';
         $secretKeyIndex = $standboxStatus ? 'test_secret_key' : 'live_secret_key';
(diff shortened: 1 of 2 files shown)


Plugin/ConfigSectionPaymentPlugin.php
```diff
@@ -69,8 +69,12 @@ private function getKeys($configData)
     {
         $configFields = $configData['groups']['omise']['fields'];
 
-        $publicKeyIndex = ($configFields['sandbox_status']) ? 'test_public_key' : 'live_public_key';
-        $secretKeyIndex = ($configFields['sandbox_status']) ? 'test_secret_key' : 'live_secret_key';
+        $standboxStatus = array_key_exists('value', $configFields['sandbox_status'])
```
hpariwat, line 72:
```suggestion
        $sandboxStatus = array_key_exists('value', $configFields['sandbox_status'])
```
I guess you meant `$sandboxStatus`.

aashishgurung: @hpariwat Yes, thanks.

kan98 (approved):

Plugin/ConfigSectionPaymentPlugin.php
```diff
@@ -69,8 +69,12 @@ private function getKeys($configData)
     {
         $configFields = $configData['groups']['omise']['fields'];
 
-        $publicKeyIndex = ($configFields['sandbox_status']) ? 'test_public_key' : 'live_public_key';
-        $secretKeyIndex = ($configFields['sandbox_status']) ? 'test_secret_key' : 'live_secret_key';
+        $sandboxStatus = array_key_exists('value', $configFields['sandbox_status'])
```
new4762, line 72: ah i see you will update it but 
i think for got about this comment 
```php
 // if sandbox status is updated the updated value will be under 'value' key else it won't have the value key
 ```

new4762 (approved):

Plugin/ConfigSectionPaymentPlugin.php
aashishgurung: @new4762 Thanks.

hpariwat (approved):

tookaa2 (approved):

PR #265, opened 2020-08-21
Hiding input card details form on selecting payment with saved card.

mayurkathale:
#### 1. Objective

This PR fixes the issue on checkout page related to card payment. When customer choses to pay with saved cards, It still shows input form for card details. This PR hides input form as soon as customer selects saved cards.

**Related information**:
Internal Ticket: https://phabricator.omise.co/T22650

#### 2. Description of change

- Updated omise-cc-form.phtml file to bind event with function
- Added new function to omise-cc-method.js

#### 3. Quality assurance

**🔧 Environments:**
- **Platform version**: Magento CE 2.3.4.
- **Omise plugin version**: Omise-Magento 2.11.
- **PHP version**: 7.2.1.

**✏️ Details:**
Note: Assuming you have saved card already in your account.

Actual behaviour:
- Add product to cart and go to checkout page
- Select Credit/Debit card as payment method
- Click on "Use a new card", This should show input form for credit card
- Click on saved card, input form for credit card still shows up.

Expected behaviour:
- Add product to cart and go to checkout page
- Select Credit/Debit card as payment method
- Click on "Use a new card", This should show input form for credit card
- Click on saved card, this should hide input form for credit card

#### 4. Impact of the change

Credit/Debit card payment should work normally

#### 5. Priority of change

Normal

#### 6. Additional Notes

NA

PR #362, opened 2022-07-05
[PLUGINS-265]: Order status history will be updated for the refund. T…

aashishgurung:
#### 1. Objective

Enhance manual sync button so it can be used to manually update order status to reflect refunds 

Jira: [#265](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-265)

#### 2. Description of change

Logic is updated to record the refunds. A new refund method is added that adds the refund order status history and updates the order status to close only if the amount is fully refunded.

#### 3. Quality assurance

- Checkout with an item
- On the Omise dashboard, refund the amount.
- On the Magento dashboard, go to the order and click on `Sync Order Stautus` button
   - If the refund is partial then it should only record the status history comment.
   - If the refund is full then it should record the status history comment and update the order status to `closed`.

**🔧 Environments:**

- PHP version: 7.4.28

### DIFF
--- a/Model/SyncStatus.php
+++ b/Model/SyncStatus.php
@@ -131,7 +131,11 @@ class SyncStatus
      */
     private function refund($order, $charge)
     {
-        if ($charge['funding_amount'] == $charge['refunded_amount']) {
+        $refundedAmount = isset($charge['refunded_amount'])
+            ? $charge['refunded_amount']
+            : $charge['refunded'];
+
+        if ($charge['funding_amount'] == $refundedAmount) {
             $order->setState(Order::STATE_CLOSED);
             $order->setStatus($order->getConfig()->getStateDefaultStatus(Order::STATE_CLOSED));
         }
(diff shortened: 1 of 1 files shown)


Model/SyncStatus.php
```diff
@@ -94,27 +94,18 @@ public function sync($order)
     /**
      * @param Order $order
      * @param array $charge
+     * @return void
      */
     private function markPaymentSuccessful($order, $charge)
     {
-        $refunded_amount = isset($charge['refunded_amount'])
-            ? $charge['refunded_amount']
-            : $charge['refunded'];
```
hpariwat, line 102: I think `refunded` is used in API version 2014-2017, so we should still support both versions.

aashishgurung: @hpariwat Thanks for letting me know. I have add it again.

kan98 (approved):

hpariwat (approved):

new4762 (approved):

PR #243, opened 2020-03-12
Implementing new payment method - Paynow QR payment (Available in Singapore only)

mayurkathale:
#### 1. Objective
Enable PayNow QR Payment support.
More info about PayNow QR Payment available [here](https://www.omise.co/paynow)

#### 2. Description of change
Added and PHP classes to support PayNow QR payment.
Modified XML files to update configuration in admin section
Added PHTML UI rendering files.
Added new email template to send order confirmation email.

UI is very simple, and flow from a user perspective is the same as in the case of other payment method.

Enable PayNow method from Admin section.
image

Checkout page showing new payment method.
Screen Shot 2020-03-12 at 09 50 58

Barcode showing on payment success page.
Screen Shot 2020-03-12 at 09 54 22

Print button allowing customers to print barcode.
Screen Shot 2020-03-12 at 09 54 50

Order confirmation email showing barcode.
Screen Shot 2020-03-12 at 09 53 38

#### 3. Quality assurance
**🔧 Environments:**
Platform version: Magento CE 1.9
Omise plugin version: Omise-Magento 2.1.
PHP version: 5.6.

**✏️ Details:**
Note: Need to use API keys from Singapore account.

Turn on a PayNow QR Payment in the admin panel, check if payment is available on the checkout page.

Turn off a PayNow QR on the admin panel, the option should not be listed on the checkout page anymore.

Make an order and successful payment with the  PayNow QR payment method.

Verify if the proper amount and payment method are displayed in Omise Dashboard.

#### 4. Impact of the change
All Payment Method available for Thai merchants should work normally Magento 1.

#### 5. Priority of change
Immediate.

#### 6. Additional Notes
- There is some common code with Tesco payment method which will be refactored in next PR.

PR #363, opened 2022-07-06
[PLUGINS-243]: Refactored the events. The order status will be update…

aashishgurung:
#### 1. Objective

Enhance webhook flow to capture `refund.create` event.

Jira: [#243](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-243)

#### 2. Description of change

The events flow is refactored to use the Magento's event manager. Event.php and Complete.php are deleted and webhook observers are added to for `charge.complete` and `refund.create` event. The order status will be updated to closed via webhook if the amount is fully refunded from the Omise dashboard but, will remaining processing if the refund is partial.

#### 3. Quality assurance

- Turn on the webhook
- Checkout with an item
- On the Omise dashboard, refund the amount.
- Go to order page in the Magento dashboard
   - If the refund is partial then it should only record the status history comment.
   - If the refund is full then it should record the status history comment and update the order status to closed.

Since we removed some files and add new files, we should test for the `charge.complete` event as well.

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.26.0
- PHP version: 7.4.28


kan98 (approved):

tookaa2 (approved):

hpariwat (approved):

tanawin-opn:
Works as expected on both auto and manual captured. 👍

PR #365, opened 2022-07-15
[PLUGINS-250] : fix trim do not accept null value in php 8

ajzkk:
#### 1. Fix trim do not accept null value in php 8

- Fix trim() do not accept null value in php 8

**Related information**:
[PLUGINS-250](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-250)

**🔧 Environments:**
- **Platform version**: Magento CE 2.4.4
- **PHP version**: >= 8.0.

### DIFF
--- a/Model/Config/Config.php
+++ b/Model/Config/Config.php
@@ -74,11 +74,12 @@ class Config
      */
     public function getValue($field, $code = self::CODE)
     {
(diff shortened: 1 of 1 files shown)


hpariwat (approved):

aashishgurung (approved):
LGTM

PR #367, opened 2022-08-02
[ENGA3-250]: upgrade omise php version to 2.15.0

ajzkk:
#### 1. Objective

Upgrade omise php version to 2.15.0

[#ENGA3-250](https://opn-ooo.atlassian.net/browse/ENGA3-250)

#### 2. Description of change

Upgrade to latest omise php version 2.15.0 which support php 8.1 and replace curl with guzzle client.

#### 3. Quality assurance

- Tested on card payment.

**🔧 Environments:**

- **Platform version**: Magento CE 2.4.4
- **Omise PHP version**: 2.15.0.
- **PHP version**: 8.1.

#### 5. Priority of change

Normal

#### 6. Additional Notes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,9 +14,9 @@
     "type": "magento2-module",
     "license": "MIT",
     "require": {
-        "php": "~5.5.0|~5.6.0|~7",
+        "php": ">=7.4",
         "magento/magento-composer-installer": ">=0.3.0",
-        "omise/omise-php": "2.13.0"
+        "omise/omise-php": "2.15.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~9.5.0",


composer.json
```diff
@@ -14,9 +14,9 @@
     "type": "magento2-module",
     "license": "MIT",
     "require": {
-        "php": "~5.5.0|~5.6.0|~7",
+        "php": ">=7.4",
```
kan98, line 17: I'm sure we have a lot of merchants below this version. do we stop supporting them altogether? I remember fixing a bug recently for a merchant on php 5.x 😅

ajzkk: I think We can create backup branch before this merge to master 
and create bug fix release (x.x.+1) from that branch if needed.
@aashishgurung What do u suggest?

aashishgurung: @kan98 Thanks for raising this concern. For the merchants using older PHP versions, we need to tell them to upgrade to at least PHP 7.4 because it would be really hard to provide support for PHP 8 and PHP 5.* at the same time as there are many deprecated features plus security reasons. We don't want to be the entry point for attackers.

But this does not mean we stop supporting them. We still have to make a plan on how will we support them. We will raise this with the team. Once again, thanks for pointing this out.

kan98: but then those merchants won't get any new payment methods 🤔 may need to check with product team if this is ok

aashishgurung: @kan98 We will discuss this with our PO and she will take it to the product team. As for now, @ajzkk will run test whether our code base supports older PHP version or not.

kan98: sounds good, thanks guys!

Issue #323, opened 2022-02-11
Magento 2.4.3 error on Checkout page.


joomlacorner:
test with Magento 2.4.3 Fresh install with Sample data on PHP 7.4 
on Checkout page will show this error.
```
1 exception(s):
Exception #0 (Magento\Framework\Exception\LocalizedException): unable to load OmiseCapabilities api

Exception #0 (Magento\Framework\Exception\LocalizedException): unable to load OmiseCapabilities api
<pre>#1 Magento\Framework\ObjectManager\Factory\Compiled->create() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:150]
#2 Magento\Framework\ObjectManager\Factory\Compiled->get() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:79]
#3 Magento\Framework\ObjectManager\Factory\Compiled->create() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:150]
#4 Magento\Framework\ObjectManager\Factory\Compiled->get() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:79]
#5 Magento\Framework\ObjectManager\Factory\Compiled->create() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:150]
#6 Magento\Framework\ObjectManager\Factory\Compiled->get() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:125]
#7 Magento\Framework\ObjectManager\Factory\Compiled->parseArray() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:86]
#8 Magento\Framework\ObjectManager\Factory\Compiled->create() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:150]
#9 Magento\Framework\ObjectManager\Factory\Compiled->get() called at [vendor/magento/framework/ObjectManager/Factory/Compiled.php:79]
#10 Magento\Framework\ObjectManager\Factory\Compiled->create() called at [vendor/magento/framework/ObjectManager/ObjectManager.php:56]
#11 Magento\Framework\ObjectManager\ObjectManager->create() called at [vendor/magento/framework/View/Element/BlockFactory.php:46]
#12 Magento\Framework\View\Element\BlockFactory->createBlock() called at [vendor/magento/framework/View/Layout/Generator/Block.php:272]
[...]

PR #376, opened 2022-08-29
[ENGA3-323]: Fix the issue of not redirecting to Thank you page

aashishgurung:
#### 1. Objective

Fix the issue of not redirecting to Thank you page after completing payment when the website has no SSL.

Jira Ticket: [#323](https://opn-ooo.atlassian.net/browse/ENGA3-323)

#### 2. Description of change

Removed the previous implementation of validating whether the REFERER is omise or not because it is not reliable as some browsers limit access to not allow HTTP_REFERER to be passed. It has been replaced with `$_SERVER['HTTP_SEC_FETCH_SITE'];` that checks whether the request is user originated or not.

References:
https://stackoverflow.com/questions/6880659/in-what-cases-will-http-referer-be-empty
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Site

#### 3. Quality assurance

Checkout with 3DS and you should see the Thank you page after completing payment.

**🔧 Environments:**

- PHP version: 8.1


Helper/OmiseHelper.php
```diff
@@ -370,12 +370,9 @@ public function isCreditCardPaymentMethod($paymentMethod)
...
-    public function validate3DSReferer()
+    public function isUserOriginated()
     {
-        $refererValue = $this->header->getHttpReferer();
-        $isProduction = strpos($refererValue, 'https://api.omise.co') === 0;
-        $isStaging = strpos($refererValue, 'https://api.staging-omise.co') === 0;
-
-        return $isProduction || $isStaging;
+        // "none" means the request is a user-originated operation
+        return 'none' === $_SERVER['HTTP_SEC_FETCH_SITE'];
```
kan98: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Site there is no support of this on Safari. should we have a fallback as well? please give it a test on the safari web browser

aashishgurung: Thanks. I will look into it.

hpariwat (approved):

Helper/OmiseHelper.php
aashishgurung: @kan98 I have re-added the logic that relies on`referer` for Safari. Even though it's less reliable, it's better than having nothing. I couldn't find other solution that would meet our objective.

tanawin-opn (approved):

kan98 (approved):

ajzkk (approved):

PR #379, opened 2022-09-13
[ENGA3-280] added pipeline for code coverage with sonarcloud

ajzkk:
#### 1. Objective

Added code coverage pipeline with sonarcloud.

**Related information**:

- [ENGA3-280](https://opn-ooo.atlassian.net/browse/ENGA3-280)
- [Sonar Cloud](https://sonarcloud.io/project/overview?id=omise_omise-magento)

#### 2. Description of change

Added code coverage pipeline with sonar cloud.

#### 3. Quality assurance

N/A

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal,

#### 6. Additional Notes

N/A


kan98 (approved):

new4762 (approved):

aashishgurung (approved):

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,43 @@
...
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v2
+          with:
+            fetch-depth: 0
+          
+        - name: Setup PHP with Xdebug
+          uses: shivammathur/setup-php@v2
+          with:
+            php-version: '8.1'
```
aashishgurung, line 24: Let's use strategy matrix and create multiple jobs for different PHP versions instead of just PHP 8.1. You can check the Omise PHP Github workflow for reference.

Reference:
- https://github.com/omise/omise-php/blob/master/.github/workflows/coding-standard-check.yml
- https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

ajzkk: @aashishgurung we have another pipeline for unit test, may be we can modify that pipeline for different php version. But for code coverage one version is enough I think.

aashishgurung: Ah I thought we are modifying the same file. So, we are running test from two jobs. Maybe we can upload the XML file to the artifacts from the other job and download that XML from the artifact in this job and upload it to SoncarCloud. This way we run tests from only one job.

Let's work on that creating a different ticket. Thanks for the good work :+1:

swathi1102 (approved):

PR #384, opened 2022-09-26
[ENGA3-446]: Issue of duplicate cancelled event has been fixed.

aashishgurung:
#### 1. Objective

Fixed the issue of duplicate cancelled event.

Jira Ticket: [#446](https://opn-ooo.atlassian.net/browse/ENGA3-446)

#### 2. Description of change

In `Offsite.php`, if the webhook is enabled then we manually add the error message to the log and message manager which shows the message to the cart page, and redirect to cart page. The process of cancelling the order and restoring the product quantity is handled by webhook.

If the webhook is not enabled then we use `LocalizedException` which handles all of the above plus cancels the order and restores the product quantity automatically.

#### 3. Quality assurance

- Turn on Webhook in Magento admin
- Setup webhook in Omise dashboard
- Open product page and look at the quantity
- Checkout with Online FPX
- On reaching payment provider page, open product page and look at the quantity. It should be down by the number of quantity in the cart.
- Cancel the payment.
- Open product page and look at the quantity. It should be restored to the original quantity.
- Open order page. It should have only one cancel event under `Comments history`.

_Before_

Screen Shot 2565-09-26 at 11 09 31

Screen Shot 2565-09-26 at 11 09 22

_After_

Screen Shot 2565-09-26 at 11 07 50

Screen Shot 2565-09-26 at 11 07 38

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.29.0
- PHP version: 8.1


view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -293,7 +293,7 @@ define(
              * @return {array}
              */
             getInstallmentTerms: function (id) {
-                const installmentBackends = checkoutConfig.installment_backends;
+                const installmentBackends = this.capabilities;
```
davidhanh, line 296: Out of curiosity, do you take into account that for the [latest version of the Omise capability API ](https://www.omise.co/capability-api?version=2019-05-29) the attribute is `payment_methods` while the older ones are named `payment_backends`

new4762: lets me help answering this one 
By default our magento will using 2017 api version following this line 
https://github.com/omise/omise-magento/blob/934ae23de1c046aecab55cf68a4685c48b11157d/Model/Omise.php#L58
so that why is will work 🙏

new4762 (approved):

ajzkk (approved):

PR #385, opened 2022-09-29
[ENGA3-497]: Fixed an issue of syncing order status from Magento after reversing a charge not updating the product's quantity.

aashishgurung:
#### 1. Objective

Fixed an issue of syncing order status from Magento after reversing a charge not updating the product's quantity.

Jira Ticket: [#497](https://opn-ooo.atlassian.net/browse/ENGA3-497)

#### 2. Description of change

Changed the implementation from just adding status comment and changing status to cancelling the order and invoice.

#### 3. Quality assurance

- See the quantity of the product before placing order
- Checkout via card
- Go to Omise dashboard and reverse the charge
- Go to Magento order page and click on `Sync Order Status`
- You should see the status set to `cancelled` and the comment history like shown in the the following picture
- Check the total quantity of the product. It should be same as before placing order.

Screen Shot 2565-09-29 at 14 46 38

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.29.0
- PHP version: 8.1

### DIFF
--- a/Model/SyncStatus.php
+++ b/Model/SyncStatus.php
@@ -178,8 +178,7 @@ class SyncStatus
                 $charge['failure_message'],
                 $charge['failure_code']
             )
-        )
-        $order->save();
+        )->save();
     }
 
     /**
--- a/Model/SyncStatus.php
+++ b/Model/SyncStatus.php
@@ -178,7 +178,7 @@ class SyncStatus
                 $charge['failure_message'],
                 $charge['failure_code']
             )
-        )->save();
+        )
         $order->save();
     }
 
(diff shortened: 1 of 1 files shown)


Model/SyncStatus.php
```diff
@@ -213,12 +213,10 @@ private function markPaymentExpired($order)
...
-
-        if ($order->getState() != Order::STATE_CANCELED) {
-            $order->setState(Order::STATE_CANCELED)->setStatus(Order::STATE_CANCELED);
-        }
-
+        $this->cancelOrderInvoice($order);
+        $order->registerCancellation(
+            __('Omise: Payment reversed. (manual sync).')
+        )->save();
         $order->save();
```
new4762, line 220: just wondering is 
```php 
        $order->registerCancellation(
            __('Omise: Payment reversed. (manual sync).')
        )->save();
```
and  
```php 
$order->save();
```  
is not the same save method right 🤔 ?

new4762 (approved):

Model/SyncStatus.php
aashishgurung: Nice catch. Thanks for pointing this out. They are the same method. I removed the duplicate `save`. Thanks.

ajzkk (approved):

PR #388, opened 2022-10-11
[ENGA3-511]: Improve the implementation to prevent the customer entering redirect URL directly in the address bar of the browser

aashishgurung:
#### 1. Objective

Improve the implementation to prevent the customer entering redirect URL directly in the address bar of the browser

Jira Ticket: [#511](https://opn-ooo.atlassian.net/browse/ENGA3-511)

#### 2. Description of change

Previously, we relied on `Referer` header for Safari browser and `HTTP_SEC_FETCH_SITE` header for other browsers to prevent user from triggering redirect URL manually. We knew Referer is not reliable as some browsers limit access to not allow HTTP_REFERER to be passed but we choose it for Safari because i does not support `HTTP_SEC_FETCH_SITE`. 

This implementation was not 100% reliable as some payment gateway sets the value of `HTTP_SEC_FETCH_SITE` to `none` instead of `cross-site`. This will result in customer being redirected to abandoned cart page even though the transaction was legit.

With this PR, we removed the HEADERs checks. We create a random token, save it to the payment of the order and pass it as the query parameters with the return URI to the gateway. Once, it is redirected back, we check whether the token in the query parameter is same as the token saved in the payment of the concerned order. If yes then we proceed with the transaction else we redirect user to empty  card page with message `Invalid token. Please contact our support if you have any questions`.

Screen Shot 2565-10-11 at 14 06 05

 
#### 3. Quality assurance

- Test the cancel and success transaction with any payment methods.
  - Examples: 3DS and internet banking

- Try entering the redirect URL to the address bar of the browser. It should redirect back to empty car page with error message `Invalid token. Please contact our support if you have any questions`.
  - 3DS: http://localhost:8200/th/omise/callback/threedsecure/
  - Offsite: http://localhost:8200/th/omise/callback/offsite/
  
Link to screencast:  https://drive.google.com/file/d/1lmHnUb62TwWvZ33zpa8GRUDsxQmd9vGp/view?usp=sharing

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.29.1
- PHP version: 8.1

### DIFF
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -33,6 +33,7 @@ use Omise\Payment\Model\Config\Cc as Config;
 use Magento\Checkout\Model\Session as CheckoutSession;
 use Psr\Log\LoggerInterface;
 use Omise\Payment\Controller\Callback\Traits\FailedChargeTrait;
+use Magento\Framework\App\Request\Http;
 
 class Offsite extends Action
 {
@@ -78,7 +79,8 @@ class Offsite extends Action
         OmiseEmailHelper $emailHelper,
         Config $config,
         CheckoutSession $checkoutSession,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        Http $request
     ) {
         parent::__construct($context);
 
@@ -90,6 +92,7 @@ class Offsite extends Action
         $this->config = $config;
         $this->checkoutSession  = $checkoutSession;
         $this->logger  = $logger;
+        $this->request = $request;
 
         $this->omise->defineUserAgent();
         $this->omise->defineApiVersion();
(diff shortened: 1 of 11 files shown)


Controller/Callback/Offsite.php
```diff
@@ -110,19 +113,30 @@ public function execute()
...
 
-        if ($order->getState() !== Order::STATE_PENDING_PAYMENT) {
-            $this->invalid($order, __('Invalid order status, cannot validate the payment. Please contact our
-            support if you have any questions.'));
+        $token = $this->request->getParam('token');
+
+        if (!$token || $payment->getAdditionalInformation('token') !== rtrim($token, "/")) {
+            $this->invalid(
+                $order,
+                __('Invalid token. Please contact our support if you have any questions.')
```
kan98, line 128: I think this message is not too meaningful to the customer trying to pay. They may be confused by what token means. Can we say something like `Unknown error. Please contact our support if you have any questions`? also not as useful but won't make the customer confused what a token is used for

kan98: or any other suggestions you might have 🙂

som-m:
👍🏻 

image

tanawin-opn (approved):

Controller/Callback/Offsite.php
aashishgurung: Yeah you are right. `Invalid URL` might be better. I will ask the team for better suggestions. Thanks.

kan98 (approved):

PR #389, opened 2022-10-20
[ENGA3-520]: Fixed the issue of FPX redirecting to cart page with invalid order status. Also added a delay in  redirect process before fetching charge status.

aashishgurung:
#### 1. Objective

Fixed the issue of FPX redirecting to cart page with invalid order status. Also added a delay in  redirect process before fetching charge status.

Jira Ticket: [#520](https://opn-ooo.atlassian.net/browse/ENGA3-520)

#### 2. Description of change

Currently our plugin checks for processing and pending order state only. This caused issue when the webhook updated the order state to failed before customer reaching the redirect URL. So, we added a check for failed order state as well for Offsite.

Plus, we added a delay of 0.5 secs in Offsite and Threedsecure redirect flow before fetching the charge. This delay will be removed once the payment flow is changed in the backend.

#### 3. Quality assurance

Configuration settings
- Set webhook to `Yes`
- Set Generate invoice at order status to Processing

Test
- Checkout with FPX
  - Choose Maybank2U
- Click cancel at Maybank2U login page.

**🔧 Environments:**

Specify the details of your test environments, including, for each, the platform version (on which the plugin was run), the Omise plugin version, and the versions of your system software such as PHP or Ruby.

- Omise plugin version: Omise-Magento 2.29.1
- PHP version: 8.1

### DIFF
--- a/Controller/Callback/Offsite.php
+++ b/Controller/Callback/Offsite.php
@@ -91,7 +91,7 @@ class Offsite extends Action
         $this->emailHelper = $emailHelper;
         $this->config = $config;
         $this->checkoutSession  = $checkoutSession;
-        $this->logger  = $logger;
+        $this->logger = $logger;
         $this->request = $request;
 
         $this->omise->defineUserAgent();
@@ -131,17 +131,19 @@ class Offsite extends Action
             return $this->redirect(self::PATH_CART);
         }
 
-        if ($order->getState() === Order::STATE_PROCESSING) {
+        $orderState = $order->getState()
+
+        if ($orderState === Order::STATE_PROCESSING) {
             return $this->redirect(self::PATH_SUCCESS);
         }
 
-        if ($order->getState() !== Order::STATE_PENDING_PAYMENT) {
+        if ($orderState !== Order::STATE_PENDING_PAYMENT && $orderState !== Order::STATE_CANCELED) {
             $this->invalid($order, __('Invalid order status, cannot validate the payment. Please contact our
             support if you have any questions.'));
 
             return $this->redirect(self::PATH_CART);
         }
-        
+
         $paymentMethod = $payment->getMethod();
 
         if (!$this->helper->isOffsitePaymentMethod($paymentMethod)) {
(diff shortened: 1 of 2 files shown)


new4762 (approved):

ajzkk (approved):

Controller/Callback/Offsite.php
```diff
@@ -163,6 +165,8 @@ public function execute()
         }
 
         try {
+            // adding delay to cover the delay in updating the charge status in the Omise backend
+            usleep(500000);
```
kan98, line 169: would it be better to set for example 3 retries? then if it's still pending then return as pending. Not sure if sleep is a good idea for async, what if it takes longer than 0.5 seconds?

Controller/Callback/Threedsecure.php
```diff
@@ -137,6 +137,8 @@ public function execute()
         }
 
         try {
+            // adding delay to cover the delay in updating the charge status in the Omise backend
+            usleep(500000);
```
kan98, line 141: same here

Controller/Callback/Offsite.php
aashishgurung: This issue is random and the delay is a temporary fix until the payment response flow is changed. We tested with 0.5 secs to 1 sec timing and selected the lowest time. Retry option would be better but it would slow thing for all the payment method.

kan98: I see, if this is a temporary fix then should be fine. Retry option would be faster though I think as we can do the first attempt immediately then retry after 0.25s and then after 0.5s. This means the longest wait time is 0.5s but it could also be resolved immediately. Either solution is ok with me since this is a temporary fix.

kan98 (approved):

som-m (approved):

Controller/Callback/Offsite.php
aashishgurung: Yeah, It would definitely be faster but the thing we do not know the expected charge status and all the status `success`, `pending`, and `failed` are valid. So we cannot end after first try as we won't know if the charge status we got is the actual status. So, we have to execute all the retry and use the last charge response.

PR #398, opened 2022-12-01
[ENGA3-618]: Credit memo will be automatically created for full refund if merchant clicks Sync status update button on order.

aashishgurung:
#### 1. Objective

Automatically create credit memo when a merchant does full refund from OPN Payments's dashboard and from Magento order page, clicks on `Sync Order Status` button.

Jira Ticket: [#618](https://opn-ooo.atlassian.net/browse/ENGA3-618)

#### 2. Description of change

Created a class `RefundSyncStatus` and moved the logic to refund and process credit memo there. Also, added a config to enable partial refund when creating credit memo from Magento admin.

#### 3. Quality assurance

- Make an order
- Capture the payment
- Do a full refund from OPN Payment dashboard
- On Magento admin, click on Sync order status button
- Go to credit memo

You should see a credit memo generated automatically.

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.29.2
- PHP version: 8.1

### DIFF
--- a/Model/RefundSyncStatus.php
+++ b/Model/RefundSyncStatus.php
@@ -53,7 +53,7 @@ class RefundSyncStatus
             $order->hasInvoices();
 
         if ($createCreditMemo) {
-            $this->createCredtMemo($order);
+            $this->createCreditMemo($order);
             $order->setState(Order::STATE_CLOSED);
             $order->setStatus($order->getConfig()->getStateDefaultStatus(Order::STATE_CLOSED));
         }
@@ -75,7 +75,7 @@ class RefundSyncStatus
      *
      * @return void
      */
-    private function createCredtMemo($order)
+    private function createCreditMemo($order)
     {
         $invoices = $order->getInvoiceCollection();
 
--- a/Model/RefundSyncStatus.php
+++ b/Model/RefundSyncStatus.php
@@ -79,7 +79,7 @@ class RefundSyncStatus
     {
         $invoices = $order->getInvoiceCollection();
 
-        foreach($invoices as $invoice) {
+        foreach ($invoices as $invoice) {
             $invoice = $this->invoice->loadByIncrementId($invoice->getIncrementId());
             $creditMemo = $this->creditMemoFactory->createByOrder($order);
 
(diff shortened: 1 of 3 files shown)


Model/RefundSyncStatus.php
```diff
@@ -0,0 +1,93 @@
...
+        $order->save();
+    }
+
+    /**
+     * @param Invoice $invoice
+     * @param Order $order
+     *
+     * @return void
+     */
+    private function createCredtMemo($order)
```
kan98, line 78:
```suggestion
    private function createCreditMemo($order)
```

etc/config.xml
```diff
@@ -24,6 +24,7 @@
                 <can_use_checkout>1</can_use_checkout>
                 <can_authorize>1</can_authorize>
                 <can_capture>1</can_capture>
+                <can_refund_partial_per_invoice>1</can_refund_partial_per_invoice>
```
kan98, line 27: have you confirmed that all these payment methods allow partial refunds? because not all our payment methods do

ajzkk (approved):

etc/config.xml
aashishgurung: The reason for adding this to all payment methods even though the payment method does not support it, is to allow merchants to change the order quantity while creating credit memo, which is used to refund order. Right now, merchants cannot change the order quantity.

For example, if a user orders 10 items but later want to refund 5 items then our plugin do not support that. Merchant would want to create a credit memo for 5 items only. But, our plugin will not allow that because the payment method do not have this partial refund config. By adding this config, we allow merchants to change the quantity so they can do partial refund

If a payment method do not support partial refund then we will receive the error from the API and show a message to the merchants telling them that the payment method do not support partial refund.

But by not adding for some payment methods, merchant won’t be able to change the quantity and might complain that s/he is unable to do partial refund, not knowing that the selected payment method do not support it.

kan98: awesome, thanks for explaining @aashishgurung. It makes sense now.

kan98 (approved):

PR #401, opened 2022-12-08
[ENGA3-618]: Issue of credit memo not creating on webhook has been fixed

aashishgurung:
#### 1. Objective

Fix the issue of credit memo being not created on webhook.

#### 2. Description of change

Change the webhook event from `charge.complete` to `charge.capture` so that webhook triggers the code that creates invoices because without invoice we can't create credit memo. We created `CreditMemoService` that abstracts the creation of credit memo from refund classes. We added the `CreditMemoService` class into the `RefundSyncStats` and `WebhookRefundObserver.php` as dependency and used its `create` method to generate credit memo.

***###Edit###***
Added back `charge.complete` as it is required for webhook flow. Now, we listen to both `charge.complete` and `charge.capture`, later for manual capture.

#### 3. Quality assurance

Check credit memo creation on webhook and Sync order status.

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.29.2
- PHP version: 8.1


Controller/Callback/Webhook.php
```diff
@@ -24,7 +24,7 @@ class Webhook extends Action
     protected $request;
 
     private $supportedEvents = [
-        'charge.complete',
+        'charge.capture',
```
ajzkk, line 27: will this event trigger for auto capture? As per [documentation](https://www.omise.co/api-webhooks#charge-events) it seem like will trigger only on manual capture.

aashishgurung: I am lookin into it. Thanks for pointing this out.

ajzkk (approved):

hpariwat (approved):

PR #404, opened 2022-12-14
[ENGA3-690]: Added ShopeePay under Thailand and Singpaore.

aashishgurung:
#### 1. Objective

Add ShopeePay for direct integration for Thailand and Singapore. If the website is opened from mobile device then ShopeePay should use `Jump App` i.e open ShopeePay app.

Jira Ticket: [#690](https://opn-ooo.atlassian.net/browse/ENGA3-690)

#### 2. Description of change

Added a function `isMobilePlatform` to check if the current platform is a mobile or not. We added `THB` and `SGD` under the supported currency for ShopeePay.

#### 3. Quality assurance

You should enabled ShopeePay direct and ShopeePay Jump App for your account from Opn Payment admin dashboard.

Source type: `Shopeepay`
- Place an order selecting ShopeePay
- You should see QR code in a separate page

Screen Shot 2565-12-14 at 16 07 40

Source type: `Shopeepay Jump App`
- Place an order selecting ShopeePay
- You should see a different page if you are in laptop that shows QR code and the process to scan using phone. 
Screen Shot 2565-12-14 at 16 08 54

**🔧 Environments:**

- Omise plugin version: Omise-Magento 2.30.0
- PHP version: 8.1


ajzkk (approved):

kan98 (approved):

Gateway/Request/APMBuilder.php
```diff
@@ -280,7 +280,7 @@ public function build(array $buildSubject)
                 break;
             case Shopeepay::CODE:
                 $paymentInfo[self::SOURCE] = [
-                    self::SOURCE_TYPE => 'shopeepay',
+                    self::SOURCE_TYPE => $this->helper->isMobilePlatform() ? 'shopeepay_jumpapp' : 'shopeepay',
```
new4762, line 283: can you make sure `shopeepay_jumpapp` is exist first bc 
not every one who have shopeepay (direct version) is will have jump app enable

tookaa2 (approved):

Gateway/Request/APMBuilder.php
```diff
@@ -280,11 +286,24 @@ public function build(array $buildSubject)
...
+    {
+        $defaultSourceType = Shopeepay::ID;
+        $isShopeepayJumpAppEnabled = $this->capabilities->isBackendEnabled(Shopeepay::JUMPAPP_ID);
+        $isShopeepayEnabled = $this->capabilities->isBackendEnabled(Shopeepay::ID);
+
+        if ($this->helper->isMobilePlatform() && $isShopeepayJumpAppEnabled) {
+            return Shopeepay::JUMPAPP_ID;
+        }
+
+        return $isShopeepayEnabled ? Shopeepay::ID : Shopeepay::JUMPAPP_ID;
```
new4762, line 307: what the use for `$defaultSourceType` na maybe we want default to be `shopeepay` right? 🤔

Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -59,9 +63,43 @@ public function getConfig()
...
+        }
+
+        $isShopeepayJumpAppEnabled = $this->capabilities->isBackendEnabled(Shopeepay::JUMPAPP_ID);
+        $isShopeepayEnabled = $this->capabilities->isBackendEnabled(Shopeepay::ID);
+
+        if ($this->helper->isMobilePlatform() && $isShopeepayJumpAppEnabled) {
+            return $jumpAppBackend;
+        }
+
+        return $isShopeepayEnabled ? $mpmBackend : $jumpAppBackend;
```
new4762, line 103: here too i think maybe we want default to be shopeepay

Gateway/Request/APMBuilder.php
aashishgurung: Ah that variable won't be needed. I forgot to remove that line. `Shopeepay::ID` value is `shopeepay`. So, we are returning `shopeepay` if it is enabled. Before that we are checking for mobile device.  I have added comments to make it clear.

Model/Ui/CapabilitiesConfigProvider.php
aashishgurung: Here we return the backend, not the source. Although the logic is same as above. If the user is using mobile and jump app is enabled then we return shopeepay jump app backend. If not then we check if Shopeepay dierct is enabled or not. If yes we return direct backend else the jump app because if `shopeepay` is not enabled then `shopeepay_jumpap` must be enabled in order for this code to get executed. If none of them is enable then this code will never execute.

new4762 (approved):

PR #427, opened 2023-04-18
Get installment minimum from capability

ajzkk:
#### 1. Objective

Get installment minimum from capability instead of hard coding .

Screenshot 2023-04-19 at 9 39 54 AM

Screenshot 2023-04-19 at 9 41 19 AM

### DIFF
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -277,6 +277,36 @@ class OmiseHelper extends AbstractHelper
     }
 
     /**
+     * @param  string  $currency
+     * @param  integer $amount
+     *
+     * @return string
+     */
+    public function currencyAmountFormat($currency, $amount)
+    {
+        $supportedCurrencies = [
+            'EUR',
+            'GBP',
+            'SGD',
+            'THB',
+            'USD',
+            'AUD',
+            'CAD',
+            'CHF',
+            'CNY',
+            'DKK',
+            'HKD',
+            'MYR',
+        ];
+
+        if (in_array($currency, $supportedCurrencies)) {
+            return $amount /= 100;
+        }
+
+        return $amount;
+    }
+
+    /**
      * Convert tesco code returned from Omise Backend in SVG format to HTML format
      *
      * @param  string  $svg
--- a/Helper/OmiseHelper.php
+++ b/Helper/OmiseHelper.php
@@ -247,66 +247,6 @@ class OmiseHelper extends AbstractHelper
     }
 
     /**
-     * @param  string  $currency
-     * @param  integer $amount
-     *
-     * @return string
-     */
-    public function omiseAmountFormat($currency, $amount)
-    {
-        $supportedCurrencies = [
-            'EUR',
-            'GBP',
-            'SGD',
-            'THB',
-            'USD',
-            'AUD',
-            'CAD',
-            'CHF',
-            'CNY',
-            'DKK',
-            'HKD',
-            'MYR',
-        ];
-
-        if (in_array($currency, $supportedCurrencies)) {
-            return $amount *= 100;
-        }
-
-        return $amount;
-    }
-
-    /**
-     * @param  string  $currency
-     * @param  integer $amount
-     *
-     * @return string
-     */
-    public function currencyAmountFormat($currency, $amount)
-    {
-        $supportedCurrencies = [
-            'EUR',
-            'GBP',
-            'SGD',
-            'THB',
-            'USD',
-            'AUD',
-            'CAD',
-            'CHF',
-            'CNY',
-            'DKK',
-            'HKD',
-            'MYR',
-        ];
-
-        if (in_array($currency, $supportedCurrencies)) {
-            return $amount /= 100;
-        }
-
-        return $amount;
-    }
-
-    /**
      * Convert tesco code returned from Omise Backend in SVG format to HTML format
      *
      * @param  string  $svg
--- a/Helper/OmiseMoney.php
+++ b/Helper/OmiseMoney.php
@@ -19,28 +19,20 @@ class OmiseMoney
 
     /**
      * @param integer|float $amount
-     * @param string $currency
+     * @param this
      */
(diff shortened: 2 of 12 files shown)


aashishgurung:
Can you please add the screenshot? I am getting the message as well as the list of bank even when the amount is below the minimum amount.

ajzkk:
> Can you please add the screenshot? I am getting the message as well as the list of bank even when the amount is below the minimum amount.

I have updated screenshots. You might need to change `omise-php` version to "dev-master". Once new version of `omise-php` is released I will update new `omise-php` version in this PR. Meanwhile I will set as draft PR.

view/frontend/web/js/view/payment/method-renderer/omise-offsite-installment-method.js
```diff
@@ -124,6 +120,15 @@ define(
                 return this;
             },
 
+            /**
+             * Get installment min amount from capability
+             *
+             * @returns {number}
+             */
+            getInstallmentMinLimit: function () {
+                return checkoutConfig.omise_installment_limit_amount.min / 100;
```
danfowler, line 129: The denominator depends on the currency. Can we do something similar to https://github.com/omise/omise-woocommerce/blob/3290c751e95a6b872c77d76cc64b5b3d26d54be2/includes/gateway/class-omise-payment-installment.php#L82

ajzkk: I have added `currencyAmountFormat` and wrap to it. 
I did not use the same name as wooCommerce since there is already `omiseAmountFormat` which do the opposite of `currencyAmountFormat`.

Helper/OmiseHelper.php
```diff
@@ -276,6 +276,36 @@ public function omiseAmountFormat($currency, $amount)
         return $amount;
     }
 
+    /**
+     * @param  string  $currency
+     * @param  integer $amount
+     *
+     * @return string
+     */
+    public function currencyAmountFormat($currency, $amount)
```
danfowler, line 285: What the purpose of this function? I'm not sure what `currencyAmountFormat` means?

Helper/OmiseHelper.php
```diff
@@ -276,6 +276,36 @@ public function omiseAmountFormat($currency, $amount)
         return $amount;
     }
 
+    /**
+     * @param  string  $currency
+     * @param  integer $amount
+     *
+     * @return string
```
danfowler, line 283: Does it really return `string`?

Helper/OmiseHelper.php
```diff
@@ -276,6 +276,36 @@ public function omiseAmountFormat($currency, $amount)
...
+     */
+    public function currencyAmountFormat($currency, $amount)
+    {
+        $supportedCurrencies = [
```
danfowler, line 287: Is this an array of `supportedCurrencies` or an array of `nonZeroDecimalCurrencies`? https://www.omise.co/currency-and-amount#units

Model/Capabilities.php
```diff
@@ -129,6 +129,17 @@ public function getTokenizationMethods()
         return $this->capabilitiesAPI->getTokenizationMethods();
     }
 
+    /**
+     * @return int
```
danfowler, line 133: What if the installment min amount provided by capability API doesn't end with `00`?

Helper/OmiseHelper.php
aashishgurung, line 287: Did you miss to add Japan currency here? This code `return $amount /= 100;` will not work for Japanese currency.

Helper/OmiseMoney.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @var string
+     */
+    private $currency;
+
+    /**
+     * @param integer|float $amount
+     * @param this
+     */
+    public function parse($amount, $currency)
```
aashishgurung, line 24: Could we please rename `parse` to some other name? `parse` is usually used for methods that take in strings and parse them into other data types. But here we are just setting values. So, maybe `setAmountAndCurrency` will be more descriptive or any other name you have in mind but not `parse`.

ajzkk: can we go with `init` or `setValues`?

danfowler: Is this not a constructor? https://www.php.net/manual/en/language.oop5.decon.php

ajzkk: Yes it can be constructor, but I am trying to avoid creating new class every times when we want to convert units.

ajzkk: It just because I can inject dependency into other dependency. Magento design is mostly dependency injection pattern.

example here.
```
 public function __construct(
        Helper $helper,
        ReturnUrlHelper $returnUrl,
        Config $config,
        Capabilities $capabilities,
        OmiseMoney $money
    ) {
        $this->helper = $helper;
        $this->returnUrl = $returnUrl;
        $this->config = $config;
        $this->capabilities = $capabilities;
        $this->money = $money;
    }
```

aashishgurung: Having no constructor does not help in unnecessary memory allocation. PHP will allocate memory for the object regardless whether the class has a constructor or not.

Personally, I would avoid setting values in constructor because the dependency is injected by the framework/platform via some configuration code and it will be difficult to change since the values would be dynamic. I would prefer setter method instead.

```
$this->money->setMoney($item['base_original_price']);
$this->money->setCurrency($currency);
$this->money->toSubunit();

// OR

$this->money->setMoneyAndCurrency($item['base_original_price'], $currency);
$this->money->toSubunit();

// OR
$this->money->setValues([
    'amount' => $item['base_original_price']],
    'currency' => $currency
]);
$this->money->toSubunit();
```

If we want to use `setValues`, i would prefer an array type so that it would be more clear what values we are setting.

aashishgurung (approved):

danfowler (approved):

PR #364, opened 2022-07-12
Added unit tests for OmiseHelper.php. Added a Github workflow to inst…

aashishgurung:
#### 1. Objective

Compute the code coverage automatically. Also, add Github workflow to run unit test on PR merge and setup to generate code coverage locally.

Jira Ticket: [#252](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-252)

#### 2. Description of change

Added a unit test for OmiseHelper.php class and added a Github workflow to run unit test on PR merge. A `phpunit.xml` configuration file is added. Previous test files under `Test/Model/Api` are moved under `archive`. We will improve those tests later. PR merge will also generate a new badge for code coverage that will be shown in the README.

Screen Shot 2565-07-22 at 17 12 04

**🔧 Environments:**

- Platform version: Magento 2.4.4
- Omise plugin version: Omise-Magento 2.26.0
- PHP version: 7.4.28

PR #428, opened 2023-04-28
Added PayPay payment method

ajzkk:
#### 1. Objective

Added PayPay payment method

#### 2. Description of change

Screenshot 2023-04-28 at 1 58 59 PM

#### 3. Quality assurance

- Change currency to JPY and add Japan PSP credentials. 
- Enable PayPay payment method.

#### 4. Impact of the change

N/a

#### 5. Priority of change

Normal


Cron/OrderSyncStatus.php
```diff
@@ -41,6 +41,7 @@ class OrderSyncStatus
         "omise_offsite_maybankqr",
         "omise_offsite_shopeepay",
         "omise_offsite_atome",
+        "omise_offsite_paypay",
```
danfowler, line 44: Why can't new payment methods register themselves?

etc/adminhtml/system.xml
```diff
@@ -920,6 +920,36 @@
                     </field>
                 </group>
 
+                <group id="omise_offsite_paypay" translate="label" sortOrder="535" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable PayPay QR Payment</label>
```
danfowler, line 924: Elsewhere we call this "PayPay Payment"

https://github.com/omise/omise-magento/blob/fcc3f8d07f1927f1cd153b8d467c1db51a65cb89/Helper/OmiseHelper.php#L206

etc/adminhtml/system.xml
```diff
@@ -920,6 +920,36 @@
...
+                    <comment>Enable customers to pay using PayPay QR payment method</comment>
+                    <field id="active" translate="label comment" type="select" sortOrder="524" showInDefault="1" showInWebsite="1" showInStore="1">
+                        <label>Enable/Disable</label>
+                        <comment>Enable PayPay QR payment method.</comment>
```
danfowler, line 928: https://github.com/omise/omise-magento/blob/fcc3f8d07f1927f1cd153b8d467c1db51a65cb89/Helper/OmiseHelper.php#L206

etc/adminhtml/system.xml
```diff
@@ -920,6 +920,36 @@
...
+                    <field id="title" translate="label" type="text" sortOrder="525" showInDefault="1" showInWebsite="1" showInStore="1">
+                        <label>Title</label>
+                        <comment>This controls the title which the user sees during checkout.</comment>
+                        <config_path>payment/omise_offsite_paypay/title</config_path>
+                    </field>
+                    <field id="allowspecific" translate="label comment" type="select" sortOrder="526" showInDefault="1" showInWebsite="1" showInStore="1">
+                        <label>Allowed Countries</label>
+                        <config_path>payment/omise_offsite_paypay/allowspecific</config_path>
+                        <source_model>Magento\Payment\Model\Config\Source\Allspecificcountries</source_model>
+                        <comment>If not set to all, guest customers will not have a billing country and may not be able to checkout.</comment>
```
danfowler, line 941:
```suggestion
                        <comment>If not set to all, guest customers will not have a billing country and may not be able to check out.</comment>
```

danfowler: I think "check out" will be the verb.

etc/adminhtml/system.xml
```diff
@@ -920,6 +920,36 @@
...
+                    </field>
+                    <field id="specificcountry" translate="label" type="multiselect" sortOrder="527" showInDefault="1" showInWebsite="1" showInStore="1">
+                        <label>Payment from Specific countries</label>
```
danfowler, line 944:
```suggestion
                        <label>Payment from specific countries</label>
```

view/frontend/layout/checkout_index_index.xml
```diff
@@ -104,6 +104,9 @@
                                                                         <item name="omise_offsite_shopeepay" xsi:type="array">
                                                                             <item name="isBillingAddressRequired" xsi:type="boolean">true</item>
                                                                         </item>
+                                                                        <item name="omise_offsite_paypay" xsi:type="array">
+                                                                            <item name="isBillingAddressRequired" xsi:type="boolean">true</item>
```
danfowler, line 108: How does this interact with the fact that on our API PayPay has billing address as optional?

ajzkk: I believe this is frontend UI config to show billing address component on checkout page.

Screenshot 2023-05-02 at 4 57 55 PM

Cron/OrderSyncStatus.php
ajzkk: We can. Original design was not designed to be able to register automatically.
For now, we have to register manually unless we change the structure.

danfowler: Great, thanks.

Test/Unit/PayPayAPMBuilderTest.php
aashishgurung, line 1: Great job. The test would be better qualified as integration test. So, it would be better if we move this test to integration folder as test is not mocking classes like `OmiseMoney` and `PaymentDataObject`.

Unit test might not be feasible for this class. Unit test should be focused on testing the behavior of a single unit of code in isolation from its dependencies which is very hard to achieve for `APMBuilderTest`.

ajzkk: If we move to new `Integration` folder, we might need to 
- modify Unit test workflow (which only run test from `Test/Unit` folder)
- modify Code coverage workflow to change path from `Test/Unit` to only `Test` 
- we also have to remove archive test in `Test/archive` folder which preventing to pass test cases.

Shall we do in this PR? @aashishgurung @danfowler

aashishgurung: Oh yeah. Let's discuss and plan to do those in later tasks. Maybe, it falls under `component` rather than `integration`.

etc/adminhtml/system.xml
```diff
@@ -920,6 +920,36 @@
                     </field>
                 </group>
 
+                <group id="omise_offsite_paypay" translate="label" sortOrder="535" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable PayPay Payment</label>
+                    <comment>Enable customers to pay using PayPay QR payment method</comment>
```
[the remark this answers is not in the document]
necessarylion: It's [PayPay payment](https://www.omise.co/paypay)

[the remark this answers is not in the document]
ajzkk: It's [PayPay payment](https://www.omise.co/paypay)

Test/Unit/PayPayAPMBuilderTest.php
danfowler: > we also have to remove archive test in Test/archive folder which preventing to pass test cases.

We have to do this anyway as there's no use having tests that aren't run. We said as much in https://github.com/omise/omise-magento/pull/364 😄

etc/adminhtml/system.xml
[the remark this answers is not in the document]
danfowler: Should we keep the naming consistent then? Even in the comment?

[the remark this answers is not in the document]
ajzkk: Ahh. I missed that one.

PR #434, opened 2023-06-02
Sub items with non zero amount being sent to Atome payment

ajzkk:
#### 1. Objective

Sub items with non zero amount being sent to Atome payment

#### 2. Description of change

Checked with `getParentItem()`. If parent item is has value, it mean it is sub product.

#### 3. Quality assurance

Tested with bundle product
- override sub product price
- normal sub product

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

### DIFF
--- a/Gateway/Request/APMBuilder.php
+++ b/Gateway/Request/APMBuilder.php
@@ -378,7 +378,7 @@ class APMBuilder implements BuilderInterface
 
         foreach ($items as $item) {
             $price = $item->getPrice();
-            // if item have parent item, it mean it's sub product
+            // if item has parent item, it mean it's sub product
             if ($item->getParentItem()) {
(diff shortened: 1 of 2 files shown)


Gateway/Request/APMBuilder.php
```diff
@@ -378,6 +378,10 @@ private function getOrderItems($order)
 
         foreach ($items as $item) {
             $price = $item->getPrice();
+            // if item has parent item, it mean it's sub product
+            if ($item->getParentItem()) {
+                continue;
+            }
```
danfowler, line 384: Maybe add a similar comment for the condition below.

danfowler (approved):

aashishgurung (approved):

PR #466, opened 2024-01-17
Add truemoney jumpapp

aashishgurung:
## Description

Add Truemoney Jumpapp

## Decision(s)

User will see a single Truemoney in the Magento payment settings. Plugin will decide whether Truemoney wallet or jumpapp to use a depending on which Truemoney is enabled in Opn account.
   
## Rollback procedure

`default rollback procedure`

### DIFF
--- a/Model/Ui/CapabilitiesConfigProvider.php
+++ b/Model/Ui/CapabilitiesConfigProvider.php
@@ -47,12 +47,8 @@ class CapabilitiesConfigProvider implements ConfigProviderInterface
         $listOfActivePaymentMethods = $this->_paymentLists->getActiveList($this->_storeManager->getStore()->getId());
         $currency = $this->_storeManager->getStore()->getCurrentCurrencyCode();
         $configs = [];
-
-        // Retrieve available backends & methods from capabilities api
-        $backends = $this->capabilities->getBackendsWithOmiseCode();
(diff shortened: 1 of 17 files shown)


Model/Ui/CapabilitiesConfigProvider.php
```diff
@@ -62,19 +59,40 @@ public function getConfig()
...
+     * filter only active backends
+     * @param $code         Payment method code
+     * @param $paymentList  Reference of the payment list
+     */
+    private function filterActiveBackends($code, &$paymentList)
+    {
+        // Retrieve available backends & methods from capabilities api
+        $backends = $this->capabilities->getBackendsWithOmiseCode();
+        $tokenization_methods = $this->capabilities->getTokenizationMethodsWithOmiseCode();
+        $backends = array_merge($backends, $tokenization_methods);
```
danfowler, line 78: @aashishgurung why shadow the `$backends` variable?

aashishgurung: Here, we are just merging  two arrays, `$backends` and `$tokenization_methods`, to `$backends`. Our capability API returns payment methods in two different object. One under payment methods and another under tokenization methods. For clarity, I will rename the variable names.

danfowler (approved):

AnasNaouchi (approved):

PR #522, opened 2026-04-22
MIT 3441 - Checkout session token creation and redirection

pratik-bharodiya:
## Description
* Checkout session token creation
* Redirection flow to checkout session page.

- Issue ID: 
-https://opn-ooo.atlassian.net/browse/MIT-3441
-https://opn-ooo.atlassian.net/browse/MIT-3442
 
## Rollback procedure
`default rollback procedure`


Helper/OmiseHelper.php
```diff
@@ -242,6 +241,42 @@ public function getConfig($fieldId)
...
+     * @param int
+     * @return string
+     */
+    public function checkoutSessionEndPoint()
+    {
+        $isSandbox = $this->config->isSandboxEnabled();
+        if ($isSandbox) {
+            return "https://checkout-page.staging-omise.co/api/sessions";
+        }
+        return "https://checkout-page.omise.co/";
```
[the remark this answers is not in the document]
rosle: Actually both modes should point to the same URL but with different keys - Sandbox mode, we should use test keys and non-Sandbox mode should use live keys 👍

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -0,0 +1,129 @@
...
+            )->toSubunit(),
+            'currency'        => $currency,
+            'order_id'        => (string) $order->getOrderIncrementId(),
+            'description'     => 'Magento Order id ' . $order->getOrderIncrementId(),
+            'payment_methods' => [$methodCode],
+            'redirect_urls'   => [
+                'complete_url' => "https://www.omise.co",
+                'cancel_url'   => "https://www.google.com",
+            ],
+            "refund_policy_link" => "https://opn.oo0/refund",
```
rosle: Do we want to keep this refund link? I think this is optional, maybe we can leave it blank?

Probably we can cross check these parameters with WooCommerce as well 👍

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -0,0 +1,129 @@
...
+    private $capability;
+
+    /**
+     * @var OmiseHelper
+     */
+    private $omiseHelper;
+
+    /**
+     * @param Omise\Payment\Model\Config\Cc $ccConfig
+     * @param Capabilities $capabilities
```
rosle: Should we update this to:

```suggestion
     * @param Capability $capability
```

Helper/OmiseHelper.php
```diff
@@ -242,6 +241,42 @@ public function getConfig($fieldId)
...
+    }
+
+    /**
+     * @var string
+     * @return string
+     */
+    public function getMethodId($code)
+    {
+        $codeToId = array_flip($this->omiseCodeByOmiseId);
```
rosle, line 278: **question:** ‏I'm not sure will this cause the issue because some IDs are mapped to the same code. 🤔

Model/Api/CheckoutSession.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Injecting dependencies
+     *
+     * @param Config $config
+     * @param APMSession $aPMSession
+     * @param OmiseHelper $omiseHelper
+     */
+    public function __construct(
+        Config $config,
+        APMSession $aPMSession,
```
rosle: **suggestion:** ‏Should we rename the var for consistency?

```suggestion
        APMSession $apmSession,
```

Model/Api/CheckoutSession.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * @param array $params
+     *
+     * @return Omise\Payment\Model\Api\Error|self
+     */
+    public function createSession($params)
+    {
+        try {
+            $endPoint = $this->omiseHelper->checkoutSessionEndPoint();
```
rosle: **suggestion:** Not sure should we rename this, is it one word 🤔 ‏

```suggestion
            $endpoint = $this->omiseHelper->checkoutSessionEndpoint();
```

Helper/OmiseHelper.php
```diff
@@ -242,6 +241,42 @@ public function getConfig($fieldId)
...
 
+    /**
+     * @param string $methodcode
+     * @return bool
+     */
+    public function isAllowUpa($methodCode)
+    {
+        $isUpaFeatureFlagEnabled = $this->config->getIsUpaFeatureFlagEnabled();
+        if ($isUpaFeatureFlagEnabled) {
+            return $this->isOffsitePaymentMethod($methodCode);
```
rosle: **question:** ‏UPA should be apply to offline payment methods, e.g. promptpay, as well, right? 💭

Gateway/Response/UPAPaymentDetailsHandler.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $payment       = SubjectReader::readPayment($handlingSubject);
+        $payment       = $payment->getPayment();
+        
+        $paymentType   = ($response['session']->object == "checkout_session") ? $response['session']->object : null;
+        $order         = $payment->getOrder();
+
+        $payment->setAdditionalInformation('upa_redirect_uri', $response['session']->redirect_url);
+        $payment->setAdditionalInformation('session_id', $response['session']->id);
+        $payment->setAdditionalInformation('payment_type', $paymentType);
```
rosle, line 37: **question:** ‏Not sure do we need to also add the expires at. UPA is also returning that data but probably we might need to recheck if that expiration is for session or charge.

etc/di.xml
```diff
@@ -3,6 +3,61 @@
...
+                <item name="offsite" xsi:type="string">Omise\Payment\Gateway\Request\APMBuilder</item>
+            </argument>
+        </arguments>
+    </virtualType>
+    <virtualType name="OmiseUPMAPMResponseHandler" type="Magento\Payment\Gateway\Response\HandlerChain">
+        <arguments>
+            <argument name="handlers" xsi:type="array">
+                <item name="paymentDetails" xsi:type="string">Omise\Payment\Gateway\Response\UPAPaymentDetailsHandler</item>
+                <item name="pendingPayment" xsi:type="string">Omise\Payment\Gateway\Response\UPAPendingPaymentHandler</item>
+                <item name="pendingInvoice" xsi:type="string">Omise\Payment\Gateway\Response\UPAPendingInvoiceHandler</item>
```
rosle: **thought:** Not sure for this pending payment and invoice, is it possible to reuse the existing one since the logic should be similar 💡 ‏

Helper/OmiseHelper.php
pratik-bharodiya: yes. right this will not create the issue. we did not modified anything inside during the UPA flow implemention.

Helper/OmiseHelper.php
pratik-bharodiya: yes i already did the change for that. but change in respective branch. let me push the same change in this branch.

etc/di.xml
pratik-bharodiya: No. we can no use existing handler for this. because both existing handler get the charge id from the response as we get session id.

Gateway/Http/Client/APMSession.php
```diff
@@ -0,0 +1,154 @@
...
+     * Creates an option for php-curl from the given request method and parameters in an associative array.
+     *
+     * @param  string $requestMethod
+     * @param  array  $params
+     *
+     * @return array
+     */
+    private function genOptions($requestMethod, $userpwd, $params, $is_json)
+    {
+        $user_agent = 'OmisePHP/' . OMISE_PHP_LIB_VERSION . ' PHP/' . PHP_VERSION;
```
rosle: Let's revisit the user agent 🤔 Since this is not a call from OmisePHP library so I think we should not put it here. This should be OmiseMagento instead. Please check `Gateway/Http/Client/APMSession.php`, maybe we could use existing user agent.

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -0,0 +1,130 @@
...
+            )->toSubunit(),
+            'currency'        => $currency,
+            'order_id'        => (string) $order->getOrderIncrementId(),
+            'description'     => 'Magento Order id ' . $order->getOrderIncrementId(),
+            'payment_methods' => [$methodCode],
+            'redirect_urls'   => [
+                'complete_url' => "https://www.omise.co",
+                'cancel_url'   => "https://www.google.com",
+            ],
+            'metadata'        => [
```
rosle, line 126: **thought:** ‏Not sure should we also have `store_id` and `store_name` as well. following the current metadata that we are attaching - https://github.com/omise/omise-magento/blob/develop/Gateway/Request/PaymentDataBuilder.php#L96-L100

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -0,0 +1,130 @@
...
+     * @return array
+     */
+    public function build(array $buildSubject)
+    {
+        $payment = SubjectReader::readPayment($buildSubject);
+        $order   = $payment->getOrder();
+        $methodCode = $payment->getPayment()->getMethod();
+        $currency = $order->getCurrencyCode();
+        
+        $methodCode = $this->omiseHelper->getMethodId($methodCode);
```
rosle: **suggestion:** ‏Should we rename this var to make it less confused, what do you think?

```suggestion
        $methodId = $this->omiseHelper->getMethodId($methodCode);
```

Helper/OmiseHelper.php
rosle: Could you help confirm the value that we sent to checkout page for mobile banking.

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -0,0 +1,130 @@
...
+            'description'     => 'Magento Order id ' . $order->getOrderIncrementId(),
+            'payment_methods' => [$methodCode],
+            'redirect_urls'   => [
+                'complete_url' => "https://www.omise.co",
+                'cancel_url'   => "https://www.google.com",
+            ],
+            'metadata'        => [
+                'order_id'  => (string) $order->getOrderIncrementId(),
+            ],
+            "enable_passkey" => true,
```
rosle: **todo:** ‏This one can be omitted and rely on the Omise account settings. 👍 

```suggestion
```

Gateway/Http/Client/APMSession.php
pratik-bharodiya: i used the our module defineUserAgent. "OmiseMagento module_version Magento magento_version"

Gateway/Response/UPAPaymentDetailsHandler.php
pratik-bharodiya: currently expire_at getting null in response because we are not passing this parameter in payload. And also this is not required because we check the payment status and updating the order based on webhook and charge_id. So i don't think so it is required.

Not pass this parameter in WooCommerce plugin as well.As i review the WooCommerce plugin payload for UPA

rosle: After checking together, this value is used for displaying promptpay QR, which in UPA flow, we won't do it. So it should be fine to not storing this expires at.

Helper/OmiseHelper.php
rosle: This one is fixed in [3e6535b](https://github.com/omise/omise-magento/pull/522/commits/3e6535b51cce5e41d26a7e8acd2e206c9805c29c)

Helper/OmiseHelper.php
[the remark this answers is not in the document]
pratik-bharodiya: As i make the url same for both production and stage. but then after stop working for stage key. so i add the condition based on the payment mode.

rosle (approved):

PR #523, opened 2026-04-22
MIT-3443 - Redirect to Order success page and update the order status

pratik-bharodiya:
## Description

* Update Order status.

- Issue ID: 
https://opn-ooo.atlassian.net/browse/MIT-3443

## Rollback procedure
`default rollback procedure`


Gateway/Response/UPAPendingPaymentHandler.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Omise\Payment\Gateway\Response;
+
+use Magento\Payment\Gateway\Response\HandlerInterface;
+use Magento\Sales\Model\Order;
+use Omise\Payment\Helper\OmiseHelper;
+
+class UPAPendingPaymentHandler implements HandlerInterface
```
rosle: **question:** ‏Did we already remove this file in the last PR?

Model/Api/CheckoutSession.php
```diff
@@ -47,11 +48,32 @@ public function __construct(
     public function createSession($params)
     {
         try {
-            $endpoint = $this->omiseHelper->checkoutSessionEndpoint();
+            $endpoint = $this->omiseHelper->checkoutSessionEndPoint();
```
rosle: **issue:** ‏Let's update the function name, I think we fixed it in the last PR.

Model/Api/CheckoutSession.php
```diff
@@ -47,11 +48,32 @@ public function __construct(
...
+    }
+
+    /**
+     * @var string
+     */
+    public function getSessionInfo($sessionId)
+    {
+        try {
+            $endpoint = $this->omiseHelper->checkoutSessionEndpoint();
+            $session = $this->apmSession->createSession(
```
rosle: **thought:** ‏Instead of reusing `createSession`, I think maybe we better create a new function `getSession` to make it clear.

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+        if (!in_array($orderState, $validOrderStates)) {
+            $this->invalid($order, __('Invalid order status, cannot validate the payment. Please contact our
+            support if you have any questions.'));
+
+            return false;
+        }
+
+        $paymentMethod = $payment->getMethod();
+
+        if (!$this->helper->isOffsitePaymentMethod($paymentMethod)) {
```
rosle: **question:** ‏Not sure do we need this check? or if needm should we include the offline methods as well?

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+        return false;
+    }
+
+    /**
+     * @return void
+     */
+    public function execute()
+    {
+        $order = $this->session->getLastRealOrder();
+        $status = $this->getRequest()->getParam('status');
```
rosle: **question:** ‏Are we using this status? If not let's remove it 👍

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+
+            if ($charge->isFailed()) {
+                return $this->handleFailure($charge);
+            }
+
+            // Do not proceed if webhook is enabled
+            if ($this->config->isWebhookEnabled()) {
+                $payment->setTransactionId($charge->id);
+                $payment->setLastTransId($charge->id);
+                $payment->setIsTransactionClosed(true);
```
rosle: **question:** ‏I didn't see we close the transaction in the existing payment flow (except refund). Should we be concern about this? 🤔

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+                    ->setFailSafe(true)
+                    ->build(Transaction::TYPE_CAPTURE);
+
+                $payment->addTransactionCommentsToOrder(
+                    $transaction,
+                    __(
+                        'Omise payment captured successfully. Charge ID: %1',
+                        $charge->id
+                    )
+                );
```
rosle: **thought:** ‏I'm actually not sure about this capture part. Could you check with the normal non-UPA flow if we add the capture comment to the order for thos offsite and offline methods?

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+            $this->invalid(
+                $order,
+                __('The URL is invalid. Please contact our support if you have any questions.')
+            );
+
+            return false;
+        }
+
+        $orderState = $order->getState();
+        $validOrderStates = [Order::STATE_PENDING_PAYMENT, Order::STATE_CANCELED, Order::STATE_PROCESSING];
```
rosle: **question:** ‏Not sure do we want to process the cancelled order here? In which case we're handling here and sy If it is acutally cancelled, does it make sense to just return back to checkout. 🤔

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,423 @@
...
+            $paymentMethod = $payment->getMethod();
+
+            if ($charge->isFailed()) {
+                return $this->handleFailure($charge);
+            }
+
+            // Do not proceed if webhook is enabled
+            if ($this->config->isWebhookEnabled()) {
+                $payment->setTransactionId($charge->id);
+                $payment->setLastTransId($charge->id);
```
rosle: Since these 2 lines are executed in both webhook and non-webhook. Does it make sense to move it before this condition? What do you think?

```php
$payment->setTransactionId($charge->id);
$payment->setLastTransId($charge->id);

if ($this->config->isWebhookEnabled()) {
```

Gateway/Response/UPAPendingPaymentHandler.php
pratik-bharodiya: Yes we already removed this file  in Prev PR ( https://github.com/omise/omise-magento/pull/522 ).

Model/Api/CheckoutSession.php
pratik-bharodiya: I used the same because code of the function remain same for both just change the parameter value. Let me give function name **processSession**. so i think this is suitable for both.

Controller/Callback/UPACallback.php
pratik-bharodiya: Actually this condition is not need. This controller call only if payment done using UPA else this controller is not called. but for safe side i added the condition for both offline and offsite.

Controller/Callback/UPACallback.php
pratik-bharodiya: Removed.

Controller/Callback/UPACallback.php
pratik-bharodiya: Yes. right. I removed the transcation closed.

Controller/Callback/UPACallback.php
pratik-bharodiya: In UPA flow we are not providing the functionality like authorize the amount and then capture. we capture the amount on placed order.So this is not requird. I removed the code.

Controller/Callback/UPACallback.php
pratik-bharodiya: Resolved

Controller/Callback/UPACallback.php
pratik-bharodiya: Yes you are right. No need to add Cancel status here. because order Cancel if customer redirect back from UPA flow or if customer Cancel the order from front-end. so No need to add Cancel status here.

Controller/Callback/UPACallback.php
rosle: If we don't need, let's remove it 👍

I guess only check `session_id` should be sufficient, right?

Controller/Callback/UPACallback.php
rosle: I think UPA supports it but I'm not sure if it is part of this first integration, Could you cross check this with WooCommerce implementation.

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,409 @@
...
+                    ->setAdditionalInformation([
+                        Transaction::RAW_DETAILS => [
+                            'omise_charge_id' => $charge->id,
+                            'status' => $charge->status
+                        ]
+                    ])
+                    ->setFailSafe(true)
+                    ->build(Transaction::TYPE_CAPTURE);
+                $payment->save();
+                $order->save();
```
rosle: **question:** ‏Few question on this

1. Could you help check is CAPTURE is proper type here 🤔 
2. Is it possible to just call `order->save()`, would that save everything? 🤔

Gateway/Http/Client/APMSession.php
rosle, line 1: **suggestion:** ‏I'm thinking about this one, it seems to be generic http request helper, right? do you think it will make sense to:

- Rename and move it to helper folder `Helper/OmiseHttpClient` - It seems we have some omise-related naming in that class, I think we can use this naming for now. But later if needed, we can refactor it to be more generic.
- Rename `processSession` to `request`

Model/Api/CheckoutSession.php
```diff
@@ -47,11 +48,32 @@ public function __construct(
...
                 $endpoint."api/sessions",
+                OmiseApiResource::REQUEST_POST,
                 $this->config->getSecretKey(),
-                $params
+                $params,
+                true
+            );
+            $this->refresh($session);
+        } catch (Exception $e) {
+            throw new LocalizedException(__('Failed to charge : ' . $e->getMessage()));
```
rosle: Let's use more specific message, maybe `Failed to create session`, so we know exactly when the error happens.

Please also do the same to getSession, maybe something like - `Failed to get session info`

Gateway/Http/Client/APMSession.php
pratik-bharodiya: I copy the processSession function into Request Helper class. and call the curl function from helper class.

Model/Api/CheckoutSession.php
pratik-bharodiya: Changed

Gateway/Response/UPAPendingPaymentHandler.php
rosle: Then I think we can safely delete this file from this PR?

Model/Api/CheckoutSession.php
pratik-bharodiya: Yes. There is difference in method name letter-case is different. So i make it same(checkoutSessionEndPoint)

Controller/Callback/UPACallback.php
pratik-bharodiya: I removed the condition. As this controller called when payment happen using UPA flow.

Controller/Callback/UPACallback.php
pratik-bharodiya: I added the comment. As i see in normal flow we add the comment for the capture.

Controller/Callback/UPACallback.php
pratik-bharodiya: Done. order save internally save the payment instance. so need to save payment separately

Controller/Payment/Cancel.php
```diff
@@ -0,0 +1,47 @@
...
+            if ($lastOrderId) {
+                $order = $this->orderRepository->get($lastOrderId);
+                if ($order->getState() !== \Magento\Sales\Model\Order::STATE_CANCELED) {
+                    $order->registerCancellation('Payment cancelled by customer (Omise cancel).');
+                    $this->orderRepository->save($order);
+                }
+            }
+        } catch (\Exception $e) {
+            $this->logger->info('Order Cancel Error: ' . $e->getMessage());
+        }finally {
```
rosle: Small code format to be improved here

```suggestion
        } finally {
```

Controller/Payment/Cancel.php
```diff
@@ -0,0 +1,47 @@
...
+    public function execute()
+    {
+        try {
+            $lastOrderId = $this->checkoutSession->getLastOrderId();
+            if ($lastOrderId) {
+                $order = $this->orderRepository->get($lastOrderId);
+                if ($order->getState() !== \Magento\Sales\Model\Order::STATE_CANCELED) {
+                    $order->registerCancellation('Payment cancelled by customer (Omise cancel).');
+                    $this->orderRepository->save($order);
+                }
```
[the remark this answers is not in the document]
pratik-bharodiya: This controller call only when customer back from the UPA checkout page and in that order status set to penddng payment. Order status processing will set only after the payment done.

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -120,8 +87,14 @@ public function build(array $buildSubject)
...
             'redirect_urls'   => [
-                'complete_url' => "https://www.omise.co",
-                'cancel_url'   => "https://www.google.com",
+                'complete_url' => $this->urlBuilder->getUrl(
+                    'omise/callback/upacallback',
+                    ['_query' => ['order_id' => (string)$order->getOrderIncrementId()]]
+                 ),
+                 'cancel_url'   => $this->urlBuilder->getUrl(
+                    'omise/payment/cancel',
+                    ['_query' => ['order_id' => (string)$order->getOrderIncrementId()]]
```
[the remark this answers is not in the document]
pratik-bharodiya: Removed parameter

Gateway/Request/UPAPaymentDataBuilder.php
```diff
@@ -120,8 +87,14 @@ public function build(array $buildSubject)
...
+                 ),
```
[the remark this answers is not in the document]
pratik-bharodiya: Resolved

Controller/Callback/UPACallback.php
pratik-bharodiya: Comment history page on normal payment flow.
000000005-Orders-Operations-Sales-Magento-Admin-24-06-2026_04_08_PM

Now comment history page look like below for only UPA flow.
000000009-Orders-Operations-Sales-Magento-Admin-24-06-2026_05_04_PM

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,410 @@
...
+        $this->omise->defineApiVersion();
+        $this->omise->defineApiKeys();
+    }
+
+    /**
+     * @var Magento\Sales\Model\Order\Payment
+     * @var string
+     * @return bool
+     */
+    private function validateChargeId($payment, $chargeId)
```
rosle: Let's move private function to the bottom of the file after protected functions

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,410 @@
...
+     */
+    private function handleSuccess($order, $chargeId, $payment, $paymentMethod)
+    {
+        // Update order state and status.
+        $order->setState(Order::STATE_PROCESSING);
+        $order->setStatus($order->getConfig()->getStateDefaultStatus(Order::STATE_PROCESSING));
+
+        $invoice = $this->helper->createInvoiceAndMarkAsPaid($order, $chargeId);
+        $this->emailHelper->sendInvoiceAndConfirmationEmails($order);
+
```
rosle: We should recheck the callback here:

- Remove order comment when webhook is enabled. Let the webhook handle it.
- Check the charge if is is captured/authorized and update the order comment accprdingly.

Controller/Payment/Cancel.php
pratik-bharodiya: Resolved.

Controller/Callback/UPACallback.php
pratik-bharodiya: Resolved

Controller/Callback/UPACallback.php
pratik-bharodiya: Resolved.

Controller/Callback/UPACallback.php
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * Mark order as success
+     *
+     * @param object $order
+     * @param string $chargeId
+     * @param object $payment
+     * @param string $paymentMethod
+     * @param object $charge
+     */
+    private function handleSuccess($order, $chargeId, $payment, $paymentMethod, $charge)
```
rosle: **suggestion:** ‏Since we need to use both id and other charge property, how about we just simply passing charge here:

```suggestion
    private function handleSuccess($order, $charge, $payment)
```

Charge ID can be get from `$charge` and same for payment method might be ok to retrieve from `$payment`. What do you think?

rosle (approved):

PR #527, opened 2026-06-16
Hot fix for broken unit-test pipeline

pratik-bharodiya:
## Description

Fix Unit-test job issue. In latest update of graycoreio/github-actions-magento2 unit-test folder was removed. So due to that unit-test workflow is breaking.
Change in workflow file and point to old version v6.0.0 of  graycoreio/github-actions-magento2
 
## Rollback procedure
`default rollback procedure`

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Create folder and file
       run: |
         mkdir -p app/etc && cd app/etc && touch vendor_path.php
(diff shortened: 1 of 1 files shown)


FhanOmise (approved):

rosle (approved):

rosle:
Not sure maybe we can also create a sync PR from master -> develop as we usually do after release. This change should also be include to develop as well if we do that. But I'm ok with this as well.

PR #531, opened 2026-07-16
Mit 3918 : Create Unit Test for Complete UPA flow

pratik-bharodiya:
## Description

Create Unit test case for UPA related files and try to cover all the scenario. Some block of code not possible to cover in unit test due to some limitation from PHPUnit side.
This PR cover the following unit test.
- UPA Payment Flow
- UPA Instalment Flow
- UPA Offsite Flow.

### Related links:

This PR cover following two JIRA Tickets. Not possible to create seperate branch as per ticket id. because both ticket are interconnected.

- Issue ID: https://opn-ooo.atlassian.net/browse/MIT-3918
- Issue ID: https://opn-ooo.atlassian.net/browse/MIT-3942

## Rollback procedure

`default rollback procedure`


Controller/Callback/UPACallback.php
```diff
@@ -151,7 +151,7 @@ public function execute()
             $paymentMethod = $payment->getMethod();
 
             if ($charge->isFailed()) {
-                return $this->handleFailure($charge);
+                $this->handleFailure($charge);
```
rosle: **question:** ‏Why do we remove this return, is it because the function throws the exception? 🤔 

I'm thinking maybe it might be better to keep return here to make sure it stops here   to prevent the unexpected flow if someone removes that exception throwing.

Model/DynamicCommandPool.php
```diff
@@ -62,8 +62,9 @@ public function get($commandCode)
             $methodCode = $quote->getPayment()->getMethod();
         }
 
-        $isWlbInstallment = $methodCode == Installment::CODE && $quote->getPayment() && $quote->getPayment()->getAdditionalInformation(InstallmentDataAssignObserver::WLB) == true ? true : false;
-
+        $isWlbInstallment = $methodCode == Installment::CODE && $quote->getPayment() && 
+            $quote->getPayment()->getAdditionalInformation(InstallmentDataAssignObserver::WLB) == true 
+            ? true : false;
```
rosle: **suggestion:** Maybe this works as the condition should already return boolean? 💡 ‏
```suggestion
        $isWlbInstallment = $methodCode == Installment::CODE && $quote->getPayment() && 
            $quote->getPayment()->getAdditionalInformation(InstallmentDataAssignObserver::WLB) == true
```

Test/Unit/Controller/Payment/CancelTest.php
```diff
@@ -0,0 +1,123 @@
...
+     * @covers \Omise\Payment\Controller\Payment\Cancel::__construct
+     */
+    public function testExecuteCancelsOrderSuccessfully()
+    {
+        $context = $this->createMock(Context::class);
+        $checkoutSession = $this->getMockBuilder(CheckoutSession::class)
+        ->disableOriginalConstructor()
+        ->onlyMethods(['restoreQuote'])
+        ->addMethods(['getLastOrderId'])
+        ->getMock();
```
rosle: **suggestion:** ‏Please help apply the proper the indent here 🙇

Test/Unit/Controller/Callback/UPACallbackTest.php
```diff
@@ -0,0 +1,1152 @@
...
+        $order = $this->createMock(Order::class);
+
+        $this->session->method('getLastRealOrder')
+            ->willReturn($order);
+
+        $order->method('getId')
+            ->willReturn(1);
+
+        $order->method('getPayment')
+            ->willReturn(null);
```
rosle: **suggestion:** ‏Can we simplify this mock setup by using `createConfiguredMock()` so the stubbed return values live alongside the mock creation?

```php
$order = $this->createConfiguredMock(Order::class, [
    'getId' => 1,
    'getPayment' => null,
]);

$this->session->method('getLastRealOrder')
    ->willReturn($order);
```

It makes the test setup a bit more compact and easier to scan.

Let's apply the same for other test cases in this file.

Controller/Callback/UPACallback.php
```diff
@@ -151,7 +151,7 @@ public function execute()
             $paymentMethod = $payment->getMethod();
```
rosle: **question:** ‏Could you help recheck if this `$paymentMethod` variable is used?

Test/Unit/Controller/Callback/UPACallbackTest.php
```diff
@@ -0,0 +1,1152 @@
...
+        $controller->method('getRequest')
+            ->willReturn($this->request);
+
+        return $controller;
+    }
+
+    /**
+     * @covers \Omise\Payment\Controller\Callback\UPACallback
+     */
+    public function testExecuteWithMissingPayment()
```
rosle: For this case, would the order note added to the order by this line: https://github.com/omise/omise-magento/blob/801b47ab7170fbedc0296fbe90509c75bcde39f6/Controller/Callback/UPACallback.php#L298 

Let's also add some assertion to make sure the correct order note is added.

rosle: **suggestion:** I think this file would be much easier to maintain if the tests were reordered a bit. Usually we should put the happy cases first followed by the unhappy error cases.

For example, the current cases could be organized like this:

1. Main charge outcomes
   - `testExecuteWithSuccessfulCharge`
   - `testExecuteWithSuccessfulAuthorizedCharge`
   - `testExecuteWithPendingCharge`
   - `testExecuteWithChargeFailure`
   - `testExecuteWithWebhookEnabledBuildsTransactionAndRedirectsToSuccess`
   - `testExecuteProcessingOrderRedirectsToSuccess`


2. Validations
   - `testExecuteWithMissingPayment`
   - `testExecuteWithInvalidOrderState`
   - `testExecuteWithoutSessionId`
   - ...

3. Exception / error handling
   - `testExecuteChargeFindThrowsException`
   - `testExecuteChargeErrorRestoresQuoteAndRedirectsToCart`


Let's also apply this same pattern to other test files 👍

Controller/Callback/UPACallback.php
rosle: **question:** ‏Just a question, for UPA, we don't want to show the payment method in the order note?

Test/Unit/Controller/Callback/UPACallbackTest.php
```diff
@@ -0,0 +1,1044 @@
...
+        $payment->expects($this->once())
+            ->method('addTransaction')
+            ->with(Transaction::TYPE_PAYMENT, $invoice, $this->anything())
+            ->willReturn($transaction);
+
+        $payment->expects($this->once())
+        ->method('addTransactionCommentsToOrder')
+        ->with(
+            $transaction,
+            $this->anything()
```
rosle: **question:** ‏For order note, is it possible to also assert the note message that we attach to the order? 🤔

pratik-bharodiya: Resolved

Controller/Callback/UPACallback.php
pratik-bharodiya: We need this comment. Suppose if web-hook is disable in that case we need to add the comment. if web-hook is enable then web-hook already doing the same.

rosle (approved):

PR #548, opened 2026-08-25
Add validate charge logic using API

pratik-bharodiya:
## Description

- Improve the logic for validate charge/payment status in callback controller. Previously callback controller work on charge_id URL parameter. In some case and specially in barcode payment method charge_id parameter not getting in callback url.  So implement charge validation using session_id.
- Remove env feature flag functionality. Now no need to set env flag in magento. UPA toggle visible to all merchant.
 
## Rollback procedure

`default rollback procedure`


Controller/Callback/UPACallback.php
```diff
@@ -26,6 +26,32 @@ class UPACallback extends Action
...
+     *
+     * @var string[]
+     */
+    private const SUCCESS_STATUSES = [
+        'successful',
+        'succeeded',
+        'completed',
+        'complete',
+        'paid',
+    ];
```
rosle: **question:** ‏Are these the valid charge/payment statuses?

Controller/Callback/UPACallback.php
```diff
@@ -118,13 +144,60 @@ public function __construct(
         $this->omise->defineApiKeys();
     }
 
+    /**
+     * @param array $payments
+     * @param string $sessionStatus
+     * @return array|null
+     */
+    private function pickPayment($payments, $sessionStatus)
```
rosle: **suggestion:** ‏Let's move this private function to the bottom.

and maybe how about rename the function to something like `getFinalPayment` ? what do you think?

Controller/Callback/UPACallback.php
```diff
@@ -137,8 +210,35 @@ public function execute()
...
+            if (empty($sessionId)) {
+                $this->cancel(
+                    $order,
+                    __('Cannot retrieve a session reference id. Please contact our support to confirm your payment.')
+                );
+                $this->checkoutSession->restoreQuote();
+                return $this->redirect(self::PATH_CART);
+            }
+            
+            $checkoutSessionInfo = $this->omiseCheckoutSession->getSessionInfo($sessionId);
```
rosle: **suggestion:** ‏Maybe we can separate this whole logic to get session payments to a new function?

`getCheckoutSession($payment)`

Controller/Callback/UPACallback.php
```diff
@@ -118,13 +144,60 @@ public function __construct(
...
+     */
+    private function pickPayment($payments, $sessionStatus)
+    {
+        if (!is_array($payments) || empty($payments)) {
+            return null;
+        }
+
+        foreach ($payments as $payment) {
+            $paymentStatus = strtolower(isset($payment['status']) ? (string) $payment['status'] : '');
+            if (!empty($payment['charge_id']) && in_array(strtolower((string) $paymentStatus), self::SUCCESS_STATUSES, true)) {
```
rosle: **thought:** Can we filter out the payment without charge ID out first so we don't have to keep checking if charge id exist in each loop 🤔 

---

1. filter out the payment without charge ID 
    -> all payments are without charge ID -> return last payment
2. filter the payment that has the same status as the session -> return payment
3. [fallback] return the last payment

Controller/Callback/UPACallback.php
```diff
@@ -137,8 +137,34 @@ public function execute()
 
         try {
             $payment = $order->getPayment();
-            $charge = $this->charge->find($chargeId);
-            if (! $charge instanceof \Omise\Payment\Model\Api\BaseObject) {
+            $checkoutSessionInfo = $this->getCheckoutSession($payment);
```
rosle: **suggestion:** ‏How about

```suggestion
            $checkoutSession = $this->getCheckoutSession($payment);
```

So the logic looks more concise

Controller/Callback/UPACallback.php
```diff
@@ -137,8 +137,34 @@ public function execute()
...
+                    $order,
+                    __('The payment session is invalid or no payment information was found. Please contact our support if you have any questions.')
+                );
+                $this->checkoutSession->restoreQuote();
+                return $this->redirect(self::PATH_CART);
+            }
+
+            $payments = $checkoutSessionInfo->payments;
+            $sessionStatus = $checkoutSessionInfo->status;
+            $terminalPayment = $this->getFinalPayment($payments, $sessionStatus);
```
rosle: **suggestion:** ‏Minor suggestion for the naming. I think we should use the same name for consistency and readability, either `terminalPayment`/`getTerminalPayment` or `finalPayment`/`getFinalPayment`.

Controller/Callback/UPACallback.php
```diff
@@ -137,8 +137,34 @@ public function execute()
...
+            if (!empty($terminalPayment) && !empty($terminalPayment['charge_id'])) {
+                $chargeId = $terminalPayment['charge_id'];
+                $charge = $this->charge->find($chargeId);
+            } else {
+                $this->invalid(
+                    $order,
+                    __('The URL is invalid. Please contact our support if you have any questions.')
+                );
+                $this->checkoutSession->restoreQuote();
+                return $this->redirect(self::PATH_CART);
```
rosle: **suggestion:** ‏Since we're doing this in many places, how about we extract it to a function. Something like- `redirectBackToCart($errorMessage)`. What do you think?

Controller/Callback/UPACallback.php
```diff
@@ -383,28 +379,49 @@ protected function cancel(Order $order, $message)
...
+        if (!is_array($payments) || empty($payments)) {
+            return null;
+        }
+
+        $paymentsWithChargeId = array_filter($payments, function ($payment) {
+            return !empty($payment['charge_id']);
+        });
+
+        if (empty($paymentsWithChargeId)) {
+            return end($payments);
```
rosle: **suggestion:** Since we cannot proceed the payment without charge ID, can we just return null from here to make it simpler? 💡  ‏

Controller/Callback/UPACallback.php
pratik-bharodiya: finalPayment is fine. because we already create the function with getFinalPayment name. so i think finalPayment is suitable.

Controller/Callback/UPACallback.php
```diff
@@ -137,8 +137,26 @@ public function execute()
...
+            $sessionStatus = $checkoutSession->status;
+            $finalPayment = $this->getFinalPayment($sessionPayments, $sessionStatus);
+
+            if (!empty($finalPayment) && !empty($finalPayment['charge_id'])) {
+                $chargeId = $finalPayment['charge_id'];
+                $charge = $this->charge->find($chargeId);
+            } else {
+                $errorMessage = __('The payment session is invalid or no payment information was found. Please contact our support if you have any questions.');
+                return $this->redirectBackToCart($order,$errorMessage);
+            }
```
rosle, line 157: How about doing early return like this I think it easier to read and consistent to other checks we had in this function.

Also we might be able to remove `!empty($finalPayment['charge_id'])` as it is always return with charge id.

```suggestion
            if (empty($finalPayment) {
               $errorMessage = __('The payment session is invalid or no payment information was found. Please contact our support if you have any questions.');
                return $this->redirectBackToCart($order,$errorMessage);
            }
             
            $chargeId = $finalPayment['charge_id'];
            $charge = $this->charge->find($chargeId);
```

rosle (approved):

